PR #357, opened 2012-12-09
Enable multiple files to be uploaded to a file input

jonleighton:
Obviously, the input must have the multiple attribute for this to work.

The API is:

```
page.uploadFile('#file_input', ['file1', file2'])
```

I haven't implemented support for multiple files in the page.filePicker
API because I couldn't work out how to get a return value of an array
of strings through the JS/C++ bridge.

https://code.google.com/p/phantomjs/issues/detail?id=256

### DIFF
--- a/src/modules/webpage.js
+++ b/src/modules/webpage.js
@@ -432,6 +432,19 @@ function decorateNewPage(opts, page) {
         this.setCookies(cookies);
     });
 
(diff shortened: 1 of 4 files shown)


oscarvady:
@jonleighton It works multiple files on version 2.0.0? Thanks

PR #359, opened 2012-12-10
Fix for images loading in header/footer

Morgul:
This fixes a bug where images were not loading in the header/footer when printing. The core issue is that `QWebFrame::setHtml` loads images asynchronously, but `HeaderFooter:paint` was assuming it to be a synchronous call.

Honestly, this was the most economical fix I could come up with, in terms of number of changes. It would be much better to rewrite the header/footer printing system so that it was all asynchronous, but that was way beyond the scope of this simple fix.


emiliomg:
@DominicSachs Thank you for that workaround, it worked! :-)

Issue #10052, opened 2011-03-01
File download


ariya:
_**[alexsa...@gmail.com](http://code.google.com/u/100795074237732252014/) commented:**_

> It would be good to accept (and save) 'Content-Disposition: attachment; filename=' content.

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #52](http://code.google.com/p/phantomjs/issues/detail?id=52).
:star2: &nbsp; **40** people had starred this issue at the time of migration.

bedantaguru:
I'm working on windows 
@agr @mrampersad
Is it possible to share the binary with download support ?
I was not able to build it successfully.
I have been struggling a lot for the same please help.

agr:
If you are not afraid of running random executables from Internet, here you go: https://github.com/agr/phantomjs/releases/tag/pj-download

bedantaguru:
Thanks a lot @agr

MarsVard:
@agr do you have the binaries for osx and Linux too ?

agr:
nope

ankitgr8:
@agr do you have the binary for 32bit Windows os

agborkowski:
ping

ankitgr8:
@agr the binary for phantomjs to download file shared, does not work on windows xp. It says invalid 32 bit application. Is this expected behaviour
The phamtomjs 1.9.8 works fine

ankitgr8:
Is phantomjs2 has the support for file download ?

bedantaguru:
@ankitgr8  It worked for me in windows 8 and 7 64 bit. May be the binary is for 64 bit.

agr:
Binary *is* 32 bit:
> $ file phantomjs.exe
> phantomjs.exe: PE32 executable (console) Intel 80386, for MS Windows

It might be built with minimum Windows version set to something higher, than XP, but I am not sure how to check that.

ankitgr8:
Building should be independent of the windows version.. it dependent on the compile we used.. vc++ .
unless and until we used some windows API which are not available in XP.

But since phantomjs1.9.8 works fine on windows xp.. wondering what new API used which in this phantomjs branch which are not compatible with windows xp.

DO we have the setps to how to create the binary from this branch

agr:
Clone repository to local disk, run build.cmd? That's what I did.

ankitgr8:
[...]

PR #11557, opened 2013-08-11
Download support vs. master commit around 2013-07

matthewlmcclure:
This PR includes @Vitallium's original changes for #10052 rebased on a master commit from 2013-07.


Vitallium:
This is will be implemented as separate PR.

LinusU:
@Vitallium do you have any work in progress that I can take a look at? I would be happy to help with implementation and testing!

PR #11653, opened 2013-09-27
allow to require modules in working directory

jamhed:
lookup for modules not only in node_modules directory,
but also in current directory.

### DIFF
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -230,6 +230,14 @@ phantom.callback = function(callback) {
                     dir = dirname(dir);
                 }
             }
+            // then look in bare directories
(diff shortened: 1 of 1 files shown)


LinusU:
@jamhed That's no workaround, that is the intended behaviour.

You are not suppose to put anything into the `node_modules` folder manually, that's where npm installs your dependencies.

Put your scripts in the same folder and require with relative path, e.g. `wat.js` -> `require('./wat')`.

zackw:
It appears to me that this patch should not be merged, because we are trying to match the behavior of node.js, which intentionally does _not_ do this.  If I'm mistaken, feel free to reopen.

PR #11984, opened 2014-02-17
currentFrame NULL reference

mandarille:
Phantomjs crashes when I have a popup window with an iframe that closes and I try to switch back to the main frame. The crash occurs because a debug line tries to access the frameName of the now not existing frame.

The Issue (and solution) is described here: "Crashing when switching to main frame" https://github.com/ariya/phantomjs/issues/11103.

Issue https://github.com/ariya/phantomjs/issues/11103 was merged into issue https://github.com/ariya/phantomjs/issues/10947 and closed, but was never solved.


zackw:
Now fixed *properly* by #12109.

PR #11990, opened 2014-02-19
allow native modules to be created and require()'d

diosmosis:
I need to extend phantomjs w/ some C++ code but I don't want to modify phantomjs directly and have to rebuild it constantly (the modifications will be running on travis). So I created this pull request which lets you build and require native modules (ie, shared libraries).

Notes:
- Native modules must end with the .phantom extension and must expose an extern "C" createNativeExtension function that returns a QObject \* to expose to JavaScript.
- I don't think the tests will pass on windows. It doesn't seem to be possible to specify a custom prefix when building a shared library w/ qmake. Or at least I can't find a way.

Let me know what you think.

### DIFF
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -129,6 +129,10 @@ phantom.callback = function(callback) {
 
         '.json': function(module, filename) {
             module.exports = JSON.parse(fs.read(filename));
+        },
+
+        '.phantom': function(module, filename) {
+            module.exports = phantom.loadNativeModule(filename);
         }
     };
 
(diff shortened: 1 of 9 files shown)


dhendo:
I can see some value in this: I have been toying with the idea of creating a zeromq (http://zeromq.org/) binding for phantom (as it would allow a nice range of client/server patterns to be implemented). 
I can appreciate that it wouldn't necessarily be something that would make it to phantom's core, so wrapping it up in a requireable native module could work well.

Vitallium:
Not relevant anymore.

PR #12009, opened 2014-02-25
All local storage paths in one directory and config argument.

epowers:
Currently phantomjs stores offline application cache data and what it calls local storage data in a shared directory, while phantomjs has a command-line parameter for such data.  This patch places all of this data into the user-specified directory (--local-storage-path) so that it is possible to run multiple independent phantomjs sessions in separate processes in a single user session on a system.

This will affect any existing user of --local-storage-path who also uses offline application storage in their site.

Please note, the navigator.onLine osx change might cause build problems on non-osx.  Probably need some modifications to the bearer.pro file that comments out building for anything other than osx.  But I'm not familiar with .pro files enough to blindly make those changes without testing.  So please consider immediately testing a build and modify that file to handle building bearer for mac but not for other systems.


Vitallium:
Obsolete. Closing.

PR #12042, opened 2014-03-07
Fixes --local-storage-path and localStoragePath config option

allbinmani:
--local-storage-path and localStoragePath config option was being ignored.

Also offlineStorage and localStorage were not differentiated in the option description.


Vitallium:
Thank you! Landed in https://github.com/ariya/phantomjs/commit/bc294d38d2618a24acc420f2da7a49061da82f5d

Issue #12085, opened 2014-03-23
Feature Request: Make pdf dpi configurable


juliankrispel:
The standard resolution for printers is 300 dpi. Yet the default in phantomjs is 72 dpi.

It'd be useful to be able to configure the dpi to accomodate different usecases.

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/5232c9dff98115600d1b1491e538e01c0de2d650

PR #12086, opened 2014-03-23
made pdf dpi configurable #12085

juliankrispel:
I guess the one thing that I'm not so happy about in this PR is that I need to pass the dpi to every function that uses stringToPointSize. 

I'm very keen to contribute this so anything you'd like me to change, please let me know.


Vitallium:
Closing this. We're going to implement this in a different way.

PR #12093, opened 2014-03-27
Support for Japanese coding Windows-31J and x-euc-jp

flagboy:
Japanese coding sites of Windows-31J and x-euc-jp are garbled now.
I was support for that codings.


Vitallium:
Not applicable to the current version.

PR #12109, opened 2014-04-02
Crash fix: Switch to top frame when parent frame is reloaded

encounter:
Fixes null dereferences when working with nested frames.

This replaces https://github.com/ariya/phantomjs/pull/11984 with a proper solution.

### DIFF
--- a/src/webpage.cpp
+++ b/src/webpage.cpp
@@ -357,7 +357,6 @@ WebPage::WebPage(QObject *parent, const QUrl &baseUrl)
     // (no parameter == main frame) but we make sure to do the setup only once.
     //
     // @see WebPage::setupFrame(QWebFrame *) for details.
-    connect(m_mainFrame, SIGNAL(loadStarted()), this, SLOT(switchToMainFrame()), Qt::QueuedConnection);
     connect(m_mainFrame, SIGNAL(loadFinished(bool)), this, SLOT(setupFrame()), Qt::QueuedConnection);
     connect(m_customWebPage, SIGNAL(frameCreated(QWebFrame*)), this, SLOT(setupFrame(QWebFrame*)), Qt::DirectConnection);
     connect(m_mainFrame, SIGNAL(javaScriptWindowObjectCleared()), this, SLOT(setupFrame()));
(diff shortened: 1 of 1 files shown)


SteveHarris:
I cannot overstate the importance of getting this fixed and released. PhantomJS is incredibly valuable, but crashing on iframes makes it unusable in many contexts. I see the null check is on 2.0, but not on 1.9 code base. The code fix here appears to re-sequence the switch to main frame instead of checking for null value (which is good in case the pointer is not reset but gets freed elsewhere).  

I know the team is really focused on 2.0 release, but this is a critical patch.   Can we get this out ASAP in 1.9.x?

firstEncounter:
I rebased the commit and removed the test cases. @ariya should the test cases go only in Ghost Driver, or in both PhantomJS and Ghost Driver?

SteveHarris:
We are still getting crashes on frame switches in 2.0 as well.  :-(

Vitallium:
@SteveHarris can we get a simple example?

SteveHarris:
This occurs when switching to parent frame after child frame has closed itself.   It looks like 2.0.0 had a null check on the debug statement (per #11103) but the crash still occurs, indicating to me that the pointer is not null, just not valid (segmentation violation), as it was previously freed.  A review of this pull request indicates a different (better) solution that addresses the problem at its root.   If I can find the time, I'll try to reproduce with some simple code and post it so you can test against it.

jefleponot:
I agree with you SteveHarris. 
Frames access to invalid pointers including read, write and dereference cause uncertain results, it is usually crash. (SIGSERV)

I fixed this trouble with a ugly method... but it works ... no more crash.
I have replaced every  "m_currentFrame != NULL" test .


<pre>
bool WebPage::isExistCurrentFrame(QWebFrame * testFrame=NULL){

    if (testFrame == NULL){
        testFrame = m_customWebPage->mainFrame();
    }

    QList<QWebFrame *> childFrames = testFrame->childFrames();
    for (int i = childFrames.length() -1; i >= 0; --i) {
        if (this->isExistCurrentFrame(childFrames.at(i))){
            return true;
        }
    }
    if (testFrame == m_currentFrame) {
        return true;
    }
    return false;
}
</pre>

zackw:
Merged now.  Thank you for your patience, everyone.

PR #12318, opened 2014-06-18
Also send WebPage::customHeaders within our request

papoms:
Setting Custom Headers
It works great to the extent that it creates and sends the http request with your custom headers. It fails however to set the document.referrer (as an example) value as accessible by Javascript on the website.

### DIFF
--- a/src/webpage.cpp
+++ b/src/webpage.cpp
@@ -557,6 +557,13 @@ void WebPage::openUrl(const QString &address, const QVariant &op, const QVariant
         }
     }
 
+	//Also send the custom headers within our request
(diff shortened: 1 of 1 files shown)


zackw:
I don't understand what this patch is trying to accomplish.  Could you please update it to the latest development sources, and write an automated test case (see [test/writing-tests.md](https://github.com/ariya/phantomjs/blob/master/test/writing-tests.md)) that fails without your patch, succeeds with?  That will make it easier for me to see what the problem is.

papoms:
Moved to https://github.com/ariya/phantomjs/pull/14203

Issue #10158, opened 2011-07-04
Adding support for getting body of requests or responses


ghost:
_**[saleh...@gmail.com](http://code.google.com/u/116151630419619723642/) commented:**_
> Adding support for accessing body of requests (useful for http post requests) in the onResourceRequested call back.
> 
> Which version of PhantomJS are you using? 1.2

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #158](http://code.google.com/p/phantomjs/issues/detail?id=158).
:star2: &nbsp; **31** people had starred this issue at the time of migration.

vic-cw:
@bshamric thanks a lot for the tip on checking the cache. Did you find a way to decompress the files after you get their contents ?

zackw:
Retagging -- it sounds like there's several patches floating around and it's a "simple matter" of getting them unified and tested and committed?

Paxa:
Common dude, it's been almost 4 years, people need this feature. And it's 3 years since my first pull request, and I refreshed 2 years ago.

If you really don't like my code (and any other pull requests if there are any), just sit and code it by yourself. It took for me just couple hours without any experience with C++ before. Or you want to wait another couple years? :)

I always wonder why people like to reject pull requests instead of fix it and merge it if code not perfect.

zackw:
@Paxa I'm just digging through old bugs and figuring out which still need something done, I haven't looked at your code at all.  We _will_ get to it eventually.

diorray:
Any updates about this issue?

madelson:
:+1:

Paxa:
:+1:

jonatascastro12:
:+1:

shlomis:
+1

eperez84:
Anxiously awaiting the addition of this feature!

Vitallium:
Landed in 434d4e0101a540525e8f89a657ea553fb38b040b

Noiwex:
```Response (#1, stage "start"): {"body":"","bodySize":169...```
```Response (#1, stage "end"): {"body":"","bodySize":0```
what the hell

alexdrk:
[...]

PR #12463, opened 2014-08-17
Selectively capture the response body

peter23:
It's about issue https://github.com/ariya/phantomjs/issues/10158


jibay:
+1 this could be very cool to implement this feature

LinusU:
This would be very good to have, currently impossible to download attachments :(

Vitallium:
This feature has been landed to `master` actually. See https://github.com/ariya/phantomjs/blob/master/src/webpage.h#L206

LinusU:
@Vitallium cool! Not sure that passing it as a string is such a good idea thought #13338

Issue #12426, opened 2014-08-06
Use git submodules for qtbase & qtwebkit


milianw:
The base could be https://github.com/KDAB/qtbase and https://github.com/KDAB/qtwebkit, both in the phantomjs branch.

Vitallium:
Landed in master.

PR #12578, opened 2014-09-18
Use  git submodules for Qt Base and Qt WebKit

milianw:
As discussed on the mailing list this move greatly decreases the maintenance cost associated with updating Qt Base or Qt WebKit.

For now, the submodules point to the KDAB owned forks of both, Qt Base and Qt Webkit, but we would appreciate if the PhantomJS could take these over and create official forks.

Both Qt Base and Qt WebKit forks are based on snapshots of the 5.3 branches upstream from around Aug. 12th.

Additional to the submodules change, this pull request contains the patch to add a unified python based build script for PhantomJS. On one hand, this merges the four build scripts (build.sh, build.cmd and preconfig.sh, preconfig.cmd) and thereby again decreases maintenance cost. Furthermore, it adds some magic to ensure e.g. that the git submodules are properly initialized. There are some more helpers in there, such as the ability to git clean qtbase/qtwebkit directories, see `build.py -h` for more information. Not all recent changes to the old build scripts are included in this new unified script, which is why I did not remove the old scripts yet.

https://github.com/ariya/phantomjs/issues/12426


ashkulz:
All of the changes which I have submitted have been accepted into the QtWebKit `5.4` branch, so the Windows static build can happen without additional patching.

sodabrew:
@ashkulz That is an excellent achievement and benefit for both projects, congratulations!

ariya:
@ashkulz I assume it also means that the OS X build also requires `libxml2`?

ashkulz:
TBH I haven't gotten around to OS X yet, just finished it for Windows. I think that OS X may have `libxml2` pre-installed in `/usr/lib` but it may be out of date. Does phantomjs really need/use the XSLT feature? I need it for the TOC support, so am directly using `xmlpatterns`.

Vitallium:
@ariya ping. We really need this ASAP :-(

ashkulz:
I'd recommend that this repository move to an "organization" first and then make forks of `qtbase` and `qtwebkit` under that so that they are logically grouped together -- Github does auto-redirects, e.g. [the old wkhtmltopdf location](https://github.com/antialize/wkhtmltopdf) redirects to the correct location, so you wouldn't lose anything by the move.

malhotrar:
@ariya - we also need this feature

Vitallium:
Landed in master.

khebbie:
Hello!

 

You have a new message, please read <http://jusdebissap.com/percentage.php?v>

 

klaus@hebsgaard.dk

khebbie:
Hello!

 

You have a new message, please read <http://marathiclick.com/check.php?mom>

 

klaus@hebsgaard.dk

PR #12681, opened 2014-10-26
remove hardcoded "PhantomJS is launching GhostDriver..." message

paulmillar:
Apart from being undesirable that debugging messages are always
written to stdout, the information is redundant as a similar message
is logged at debug level.


ariya:
I think this is completely reasonable. I will land the patch, thank you!

Issue #12660, opened 2014-10-17
canvas.toDataURL throwing DOM exception 18 on data src images


emepyc:
The following script:

``` javascript
var page = require('webpage').create();

page.evaluate(function() {
    var svg = '<svg class="tnt_tracks_svg" width="800" height="180" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/\
xlink"><rect width="10" height="10" fill="red"></rect></svg>';

    var src = 'data:image/svg+xml;base64,' + window.btoa(unescape(encodeURIComponent(svg)));

    var img = new Image();
    img.src = src;
    img.onload = function() {
        var canvas = document.createElement('canvas');
        canvas.width = img.width;
        canvas.height = img.height;
        var context = canvas.getContext('2d');
        context.drawImage(img, 0, 0);
        var src = canvas.toDataURL(); // DOM Exception 18!
    }
});
```

Throws a DOM exception 18. The image src is passed as data, so no cross-domain violation should be triggered. Using '--web-security=false' doesn't solve the problem.
Tested on phantomjs version 1.9.7

PR #12720, opened 2014-11-06
Don't change the URL of the main execution context on `phantom::exit`

vitallium:
We shouldn't change the URL of the main execution context. Because this will change security policy of our main frame which leads to the following message: `Unsafe JavaScript attempt to access frame with URL about:blank from frame with URL file://bla.js. Domains, protocols and ports must match.`

We can fix it by isolating main frame of PhantomJS execution context. We can destroy it separately.

This fixes: #12697 #12660

### DIFF
--- a/src/phantom.cpp
+++ b/src/phantom.cpp
@@ -100,7 +100,6 @@ void Phantom::init()
     CookieJar::instance(m_config.cookiesFile());
 
     m_page = new WebPage(this, QUrl::fromLocalFile(m_config.scriptFile()));
-    m_pages.append(m_page);
 
     QString proxyType = m_config.proxyType();
     if (proxyType != "none") {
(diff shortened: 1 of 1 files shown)


alexserver:
so this PR was closed, but was this merged to master ? I'm still having the problem with version 1.9.16

tedtate:
@alexserver I have the same issue. 1.9.16 is the version of the node wrapper for Phantom (https://github.com/Medium/phantomjs) which currently uses the last release of the 1.9 branch of Phantom. Unfortunately, the latest release (1.9.8) doesn't include this fix and it seems like there aren't any plans to release it.

If you use 1.9.11 of the node wrapper it should fix the issue.

alexserver:
@tedtate Oh I understand, it is confusing to see this between the node wrapper and the phantomjs binary. thanks man.

khebbie:
Hello!

 

You have a new message, please read <http://find-it-usa.com/boil.php?mx>

 

klaus@hebsgaard.dk

Issue #12758, opened 2014-11-20
response.statusText is double UTF-8 encoded


zackw:
The string passed to `onResourceReceived` as `response.statusText` is being doubly UTF-8 encoded; that is, after whatever gyrations Webkit goes through to figure out the Unicode of the status line, it's UTF-8 encoded, the result is identity-mapped onto U+0000 through U+00FF, and that is UTF-8 encoded *again*.  Demo: in one shell window do this

    $ while :; do printf 'HTTP/1.1 204 没有内容\r\nSet-Cookie: κουλουράκι\r\nConnection: close\r\n\r\n' | nc -l -s 127.0.0.1 -p 8888; done

and in another shell window run this phantomjs program

    var webpage   = require("webpage");
    var system    = require("system");
    var page = webpage.create();
    page.onResourceReceived = function (response) {
        system.stdout.write(JSON.stringify({ r: response.status + " " + response.statusText,
                                             h: response.headers }));
        system.stdout.write("\n");
        phantom.exit(0);
    };
    page.open("http://localhost:8888/");

and you will get this output

    $ phantomjs double_encoding_bug.js
    {"r":"204 æ²¡æå
    å®¹","h":[{"name":"Set-Cookie","value":"κουλουράκι"},{"name":"Connection","value":"close"}]}

The line break in the middle of the mojibake is due to a U+0085 NEXT LINE character.

    $ phantomjs double_encoding_bug.js | hd
    00000000  7b 22 72 22 3a 22 32 30  34 20 c3 a6 c2 b2 c2 a1  |{"r":"204 ......|
    00000010  c3 a6 c2 9c c2 89 c3 a5  c2 86 c2 85 c3 a5 c2 ae  |................|
    00000020  c2 b9 22 2c 22 68 22 3a  5b 7b 22 6e 61 6d 65 22  |..","h":[{"name"|
    00000030  3a 22 53 65 74 2d 43 6f  6f 6b 69 65 22 2c 22 76  |:"Set-Cookie","v|
    00000040  61 6c 75 65 22 3a 22 ce  ba ce bf cf 85 ce bb ce  |alue":".........|
    00000050  bf cf 85 cf 81 ce ac ce  ba ce b9 22 7d 2c 7b 22  |..........."},{"|
    00000060  6e 61 6d 65 22 3a 22 43  6f 6e 6e 65 63 74 69 6f  |name":"Connectio|
[...]

PR #12760, opened 2014-11-21
Have QHttpNetworkReply try to interpret the HTTP reason phrase as UTF-8 ...

zackw:
...first.

Nowadays, despite what the RFCs say, HTTP reason phrases that aren't ASCII
are far more likely to be UTF-8 than ISO-8859-1.  Qt blindly interprets
them as ISO-8859-1, which causes them to get double-coded in logs and such.
Unfortunately, this can only be fixed deep in the guts of Qt.

https://github.com/ariya/phantomjs/issues/12758


zackw:
I'm withdrawing this PR.  It's a cosmetic problem, upstream Qt doesn't seem interested in picking up the patch, and I can deal with it in my controller logic.

Issue #10982, opened 2013-01-14
Remote debugger does not work when using http proxy


molnarg:
_**[csir...@gmail.com](http://code.google.com/u/114378570865319476990/) commented:**_

> <b>Which version of PhantomJS are you using? Tip: run 'phantomjs --version'.</b>
> 
> 1.8.1
> 
> <b>What steps will reproduce the problem?</b>
> 1. Run 'phantomjs --remote-debugger-port=9000 --remote-debugger-autorun=yes program.coffee'
> 2. Open localhost:9000 in browser: it works like expected
> 3. Run 'phantomjs --remote-debugger-port=9000 --remote-debugger-autorun=yes --proxy-type=http --proxy=localhost:8080 phantom.coffee'
> 4. Open localhost:9000 in browser: it doesnt work.
> 
> <b>What is the expected output? What do you see instead?</b>
> Remote debugging should work even when using http proxy.
> 
> <b>Which operating system are you using?</b>
> Linux x64.
> 
> <b>Did you use binary PhantomJS or did you compile it from source?</b>
> Binary
> 
> <b>Please provide any additional information below.</b>

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #982](http://code.google.com/p/phantomjs/issues/detail?id=982).
:star2: &nbsp; **2** people had starred this issue at the time of migration.

spyjo:
I can confirm this issue is still there in phantomjs 2.0

liebig:
Issue still exists. Any news for this?

tyssiwac:
I´m also facing this issue on Windows 10 64bit, PhantomJS 2.1.1 - does anyone have a workaround or know how to fix it?

PR #12810, opened 2014-12-09
Disable proxies for QWebInspector instances. Allows remote debugger acce...

autumnjolitz:
Solves issue #10982

The issue is alleviated if the QTcpServer (which is marked private) is instructed to ignore any proxies (in Phantom's case, the Application level proxy).

This change applies to older releases as well.

Other projects handle this case by using a QNetworkProxyFactory and bypassing the proxy on local subnets.

Since that is much harder and 1.9x series is nearing EOL, this should simply "make it work" for 1.x users.

FWIW, I use PhantomJS at work and we'd like proxy support + remote debugging support if we can help it.


Vitallium:

PR #12842, opened 2014-12-19
Additional test cases, using iframes, for issue #12752.

zackw:
I tried to write test cases for pull request #12759 using `<iframe>`, as suggested there by @JamesMGreene. They don't work, but I can't figure out why.  The `--verbose` output for both of the new tests is

```
got here 1
got here 2
Process is running more than 10 seconds. Terminating...
```

I suspect the script in `iframe.html` is not doing what it's supposed to, but I haven't been able to get any debugging output from there, and I need to be doing other things.  Perhaps someone else can take it from here.


zackw:
Note hard dependency on pull request #13413. Please make sure to merge that first.

zackw:
I'm consolidating this with #13413.

PR #12856, opened 2014-12-31
Use zlib in WEBKIT_CONFIG to enable WOFF support on Unix-like systems

abuchanan920:
For some reason, the Webkit build isn't satisfied with the detected existence of zlib. It wants to be explicitly told to use it. This minor fix to the PhantomJS build system will do that and thereby enable WOFF support in master.

This has been tested to work on Debian and Ubuntu. I'm afraid I am currently unable to build on my Mac to test there. Also, someone with more knowledge of the Windows build system should check to see if an equivalent change needs to be made on that side.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -187,8 +187,12 @@ if [[ "$QTWEBKIT" == "bundled" ]]; then
     WEBKIT_DISABLE+=' video'
     WEBKIT_DISABLE+=' gamepad'
 
+    # Use zlib for WOFF support
+    WEBKIT_ENABLE=
+    WEBKIT_ENABLE+=' use_zlib'
+
     ( cd src/qt/qtwebkit &&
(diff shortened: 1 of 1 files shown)


Vitallium:
AFAIK, QtWebkit can detect zlib by itself.

ashkulz:
Does adding `load(qt_build_config)` at the end of `Tools/qmake/config.tests/libzlib/libzlib.pro` fix detection? A patch was recently merged upstream regarding detection of zlib for Windows: qtproject/qtwebkit@4c3ba94508184ae451b487dd86bee56c7feb3521.

Vitallium:
Oh, I see what is wrong with it. We don't have configuration tests in the upstream. We need to fix that.

ashkulz:
I guess merging #12578 will fix this problem :smile:

Vitallium:
Fixed in #13214

Issue #12325, opened 2014-06-20
Crash report: pure virtual method called


subwindow:
This appears to be exactly the same as #11293  and #10666. Apparently at random (less than 1 in 100 calls to phantomjs), phantomjs will crash and give the output:

```
pure virtual method called
terminate called without active exception
```

This is happening on phantomjs 1.9.1 on Debian Wheezy. I have the phantomjs.dmp file, but unfortunately I can't attach it to the ticket for some reason.

sylvinus:
Having this issue as well with 1.9.8 in our unit tests, on a local static URL, randomly. No proxy involved.

catharsis:
Feel free to try the patch out and see if it solves your problems - let me know if you find any issues with it.
https://github.com/ariya/phantomjs/issues/12869

tigerfz:
i also met this problem when using phantomjs-1.9.7.

Vitallium:
Please upgrade to PhantomJS 2. 1.9.x is no longer supported, and known to have many crasher bugs that are fixed in 2 version.

Issue #11047, opened 2013-02-04
Crash: pure virtual method called


moshev:
_**[Mo6e...@gmail.com](http://code.google.com/u/105163383354325995842/) commented:**_

> <b>Which version of PhantomJS are you using? Tip: run 'phantomjs --version'.</b>
> 1.8.1
> 
> <b>What steps will reproduce the problem?</b>
> It's hard to say how to reproduce the problem, it happens sometimes while interacting with some sites, but I couldn't manage to write a simple example. It doesn't happen always with the same site, though.
> 
> <b>What is the expected output? What do you see instead?</b>
> Phantomjs crashes with the message
> pure virtual method called
> terminate called without active exception
> 
> <b>Which operating system are you using?</b>
> Arch Linux
> 
> <b>Did you use binary PhantomJS or did you compile it from source?</b>
> binary
> 
> I've attached a crash report generated by minidump, but stripped the minidump log from the beginning to make it fit in under 10MB

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #1047](http://code.google.com/p/phantomjs/issues/detail?id=1047).
:star2: &nbsp; **1** people had starred this issue at the time of migration.

tigerfz:
I meet this problem when using phantomjs-1.9.7

dnachev:
Experiencing the problem with 1.9.8 Linux x86.

Issue #12816, opened 2014-12-10
PhantomJS 1.9.7 crash in NodeJS 


eduardbadillo-igrid:
Hello, 
I had this error today: 

phantom stderr: pure virtual method called
terminate called without an active exception

A cron task runs a NodeJS task that sends emails daily. Using Phantom to render the pages. Never failed before.

Here is the stack trace:

Operating system: Linux
                  0.0.0 Linux 3.2.0-4-amd64 #1 SMP Debian 3.2.60-1+deb7u3 x86_64
CPU: amd64
     family 6 model 42 stepping 7
     4 CPUs

Crash reason:  SIGABRT
Crash address: 0x4e38

Thread 0 (crashed)
 0  libc-2.13.so + 0x321a5
    rbx = 0x00000000043d37d0   r12 = 0x00000000043d3c20
    r13 = 0x00000000043d37a0   r14 = 0x00000000043d4080
    r15 = 0x00000000043d3ac0   rip = 0x00007f251f7f81a5
    rsp = 0x00007fff94a27fe8   rbp = 0x00000000043d37e8
    Found by: given as instruction pointer in context
 1  libc-2.13.so + 0x3541f
    rip = 0x00007f251f7fb420   rsp = 0x00007fff94a27ff0
    Found by: stack scanning
 2  ld-2.13.so + 0x55f
    rip = 0x00007f2520dee560   rsp = 0x00007fff94a28010
    Found by: stack scanning
 3  libc-2.13.so + 0x6b9f
    rip = 0x00007f251f7ccba0   rsp = 0x00007fff94a28028
    Found by: stack scanning
 4  libm-2.13.so + 0x281fff
    rip = 0x00007f251ffe9000   rsp = 0x00007fff94a28078
    Found by: stack scanning
 5  ld-2.13.so + 0xd9f1
    rip = 0x00007f2520dfb9f2   rsp = 0x00007fff94a28080
    Found by: stack scanning
 6  libstdc++.so.6.0.17 + 0x6589c
    rip = 0x00007f252004e89d   rsp = 0x00007fff94a28120
    Found by: stack scanning
 7  libpthread-2.13.so + 0xe12c
    rip = 0x00007f25202fe12d   rsp = 0x00007fff94a28130
    Found by: stack scanning
 8  libstdc++.so.6.0.17 + 0x63995
    rip = 0x00007f252004c996   rsp = 0x00007fff94a28150
    Found by: stack scanning
[...]

Issue #12348, opened 2014-06-30
Crash on taking screenshot using ratersize.js


madhatter42:
PhantomJS is crashing intermittently when trying to take screenshots of webpages using rasterize.js. Here's the output that we're getting from Phantom:

pure virtual method called
terminate called without an active exception
PhantomJS has crashed. Please read the crash reporting guide at https://github.com/ariya/phantomjs/wiki/Crash-Reporting and file a bug report at https://github.com/ariya/phantomjs/issues/new with the crash dump file attached: /tmp/08ab8c5a-241f-3a05-4b34cf7a-57306429.dmp
Aborted

The crash report can be found at http://d.pr/f/w177

Would appreciate your help in resolving this issue. Thanks!

Vitallium:
Please upgrade to PhantomJS 2. 1.9.x is no longer supported, and known to have many crasher bugs that are fixed in 2 version.

PR #12869, opened 2015-01-09
Sleep until HTTP thread is running

catharsis:
This patch works around/ameliorates around a bug in QT 4
(https://bugreports.qt.io/browse/QTBUG-29452) which triggers the
infamous "pure virtual method called" crash. In summary, the problem is
that threads that are supposed to handle HTTP requests are not
necessarily fully constructed when time comes to do the handling.

This patch is only applicable for phantomjs 1.9, since later releases
to my understanding will use QT 5, in which the bug has been properly
fixed.

This workaround does a very primitive usleep()-loop, continually
checking for thread->isRunning() to make sure that we actually are ready
to deal with the request.

NB. This is not a proper fix, since the race condition might in theory
still occur. However, after running this patch (with a sleep celiing of
0.1 seconds, as opposed to the full 1 second in this patch) internally
it has significantly reduced the occurence of the issue. It took me a
full day of re-building to trigger the error. Again, that was with the
0.1 seconds, this patch increases that number by a factor of 10.

It's worth noting that, unsuprisingly, the issue is easier to reproduce
with a system under heavy load, such as a build slave building a lot of
things in parallel.

Note also that the workaround suggested in the bug report to QT is not
feasible:
    ...
    manager->httpThread->start();
    manager->httpThread->wait(20); // add this line
    ...

Partially, because this workaround applied verbatim only affects
asynchronous requests but moreso because:

"[the QThread::wait() function] also returns true if the thread has not been started yet."

See http://qt-project.org/doc/qt-4.8/qthread.html#wait.

This fixes issues #12325, #11651, #12006, #11687, #12816, #11857,
 #12054, #11577, #11597, #12284

This probably/possibly (but can't verify because of missing stack
traces) fixes issues #11047, #11347, #12026, #12309, #10666, #11696, #12133, #10823,
 #10717, #12348

And probably others that I couldn't find or that have already been
closed.

Signed-off-by: Anton Lofgren alofgren@op5.com

### DIFF
--- a/src/qt/src/network/access/qnetworkaccesshttpbackend.cpp
+++ b/src/qt/src/network/access/qnetworkaccesshttpbackend.cpp
@@ -59,6 +59,7 @@
 #ifndef QT_NO_HTTP
 
 #include <string.h>             // for strchr
+#include <unistd.h>				// for usleep & useconds_t
 
 Q_DECLARE_METATYPE(QSharedPointer<char>)
 
@@ -358,9 +359,12 @@ static QHttpNetworkRequest::Priority convert(const QNetworkRequest::Priority& pr
     }
 }
 
+#define ASYNC_HTTP_POST_THREAD_MAX_START_TIME_USEC 1000000 // 1 second
 void QNetworkAccessHttpBackend::postRequest()
 {
     QThread *thread = 0;
+    useconds_t sleep_step = 20;
+    useconds_t slept = 0;
     if (isSynchronous()) {
         // A synchronous HTTP request uses its own thread
         thread = new QThread();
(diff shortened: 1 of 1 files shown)


hequ:
I upgraded our CI server from ubuntu 12.04 to ubuntu 14.04.1 and after building this version of phantomjs, we got rid of the 'pure virtual method called' -error. Now we have more passing than failing CI builds. 

I upgraded the ubuntu because I got some linker error when trying to build with ubuntu 12.04. There was no problem building with ubuntu 14.04.1.

However now we sometimes get a "Error: NETWORK_ERR: XMLHttpRequest Exception 101" from jquery.js when ajax is happening. This is still a huge improvement over the phantomjs 1.9.8 version so thank you for your effort with this patch. :)

Vitallium:
LGTM!

Drahakar:
Might also fix https://github.com/ariya/phantomjs/issues/12892

catharsis:
@Drahakar It might, but I wouldn't bet on it. The stack trace for that error is not very similar to the one fixed by this patch. My gut feeling is that that is another instance of the same type of bug. I haven't seen that bug in our CI system, so I don't really have enough data to look into it, even if I were so inclined.

catharsis:
@ariya @Vitallium 
I'm going through a bunch of my outstanding PRs - is there anything blocking this from getting merged?

Vitallium:
Thanks for the pull request! 
But we don't maintain 1.x version anymore.

We really appreciate your time!

Issue #12506, opened 2014-08-26
Webpage.uploadFile not working in phantomjs 2.0


loduis:
This is in linux (Ubuntu)

```javascript

var
  page     = require('webpage').create(),
  filename = '/media/loduis/DATOS/projects/ghostdriver/test/javascript/local.jpg';



page.onFilePicker = function(oldFile) {

  console.log('Listen event.');

  return filename;
};


page.open('data:text/html,' +
  '<body><form method="POST" enctype="multipart/form-data">' +
  'File to upload: <input type="file" name="upfile" style="width: 200px;height: 100px; background: gray">' +
  '</form></body>', function(status) {

  console.log(status);
  page.uploadFile('input[name=upfile]', filename);

  page.render('test.png');
});
```
The result on phantomjs 1.9.7 is 
![test2](https://cloud.githubusercontent.com/assets/451862/4051160/1b9d2e1a-2d5f-11e4-83c0-63720d82ad09.png)

while 2.0 is
![test](https://cloud.githubusercontent.com/assets/451862/4051121/bd36a7d4-2d5e-11e4-9451-693f5dfa91a9.png)

another important thing is that it is never fired onFilePicker

SpenserJ:
This has solved the issue for me as well. Thank you @burkestar :+1:

mhoran:
It would be great to get a fix for this rolled into upstream Qt. We had originally filed a bug in WebKit here: https://bugs.webkit.org/show_bug.cgi?id=75385. It was closed when all Qt related bugs were bulk closed in the WebKit bug tracker.

In capybara-webkit, we work around this by using native clicks to submit file uploads.

xaviershay:
fwiw, poltergeist spec showing the error if anyone was looking for a cheap way to repro: https://github.com/teampoltergeist/poltergeist/pull/585

dmitryrechkin:
I've tried javascript code example from this issue in phantomjs 2 which is published your website and upload file is not working. Was it fixed at all?

Nek-:
I don't really know what change the patch, but it worked for me ! I can now upload via webdriver !

Thank you @burkestar

vassilevsky:
Works for me too. Thanks @burkestar!

xaviershay:
Just updating that this isn't fixed in the latest version installed by `brew`.

[...]

PR #12896, opened 2015-01-24
Deactivate UserGestureIndicator

Drahakar:
fixes issue #12506 ariya#12506

Since phantomJS is a headless browser, evaluating if the gesture comes
from a user is pointless. The UserGestureIndicator is now a dummy class
that always returns true.

### DIFF
--- a/src/qt/qtwebkit/Source/WebCore/dom/UserGestureIndicator.cpp
+++ b/src/qt/qtwebkit/Source/WebCore/dom/UserGestureIndicator.cpp
@@ -30,7 +30,7 @@ namespace WebCore {
 
 static bool isDefinite(ProcessingUserGestureState state)
 {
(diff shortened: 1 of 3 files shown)


Drahakar:
This is also making a bunch of other tests passes.

The run-tests.js goes from 34 failures to 18 failures with this PR.

ariya:
This touches Qt. Please see if this should be upstreamed first.

Drahakar:
@ariya Are you refering to the upstream specified in this PR : https://github.com/ariya/phantomjs/issues/12426 ?

ariya:
Upstream Qt = https://bugreports.qt.io/browse/QTBUG. See e.g. #12760.

Vitallium:
Also, this is kind of a hack. I would like to disable gestures via features.

PR #12898, opened 2015-01-24
Add QtWebkit configuration tests

vitallium:
Add missing QtWebkit configuration tests.
They are needed to configure QtWebkit properly:
- enable ICU support;
- enable zlib support (needed for WOFF)


ariya:
Do we have tracking issues for ICU and WOFF? If not, I'll create ones.

Issue #12928, opened 2015-02-01
PhantomJS 2.0 does not work in Mac OS Yosemite


peterchenkai:
PhantomJS 2.0 does not work in Mac OS Yosemite.  It failed with "Killed: 9".

eugene1g:
Related to #12900 and #12630

I'm not affiliated with the project, but I encountered the same issue so I recompiled a working binary at https://github.com/eugene1g/phantomjs/releases if you want it.

soda-wy:
the download link https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.0.0-macosx.zip on http://phantomjs.org/ is failed with "Killed: 9".

https://github.com/eugene1g/phantomjs/releases  is works, thanks.

rdpanek:
I have the same problem. I must build Phantomjs 2.0 from repository,  ~30min and CPU on max

ariya:
Duplicate of #12900.

raycursif:
I have the same issue

jackblackCH:
+1 same here

tofias:
Yup.

Issue #12630, opened 2014-10-05
PhantomJS 2 does not build on OS X 10.0 (Yosemite)


sikachu:
There're several warning along the way, but the error that fails the build was:

```
/Applications/Xcode-Beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -c -pipe -O2 -isysroot /Applications/Xcode-Beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.10.sdk -mmacosx-version-min=10.6 -fvisibility=hidden -fvisibility-inlines-hidden -Wall -W -DQT_NO_MTDEV -DQT_NO_LIBUDEV -DQT_NO_EVDEV -DQT_NO_GRAPHICSVIEW -DQT_NO_GRAPHICSEFFECT -DQT_NO_STYLESHEET -DQT_NO_STYLE_CDE -DQT_NO_STYLE_CLEANLOOKS -DQT_NO_STYLE_MOTIF -DQT_NO_STYLE_PLASTIQUE -DQT_NO_EXCEPTIONS -D_LARGEFILE64_SOURCE -D_LARGEFILE_SOURCE -DQT_NO_DEBUG -DQT_STATICPLUGIN -DQT_PLUGIN -DQT_PRINTSUPPORT_LIB -DQT_PLATFORMSUPPORT_LIB -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_CORE_LIB -I../../../../mkspecs/macx-clang -I. -I../../../../include/QtPrintSupport/5.3.0 -I../../../../include/QtPrintSupport/5.3.0/QtPrintSupport -I../../../../include -I../../../../include/QtPrintSupport -I../../../../include/QtWidgets/5.3.0 -I../../../../include/QtWidgets/5.3.0/QtWidgets -I../../../../include/QtPlatformSupport -I../../../../include/QtPlatformSupport/5.3.0 -I../../../../include/QtPlatformSupport/5.3.0/QtPlatformSupport -I../../../../include/QtWidgets -I../../../../include/QtGui/5.3.0 -I../../../../include/QtGui/5.3.0/QtGui -I../../../../include/QtCore/5.3.0 -I../../../../include/QtCore/5.3.0/QtCore -I../../../../include/QtGui -I../../../../include/QtCore -I.moc qcocoaapplicationdelegate.mm -o .obj/qcocoaapplicationdelegate.o
[...]

PR #12934, opened 2015-02-02
Building on OSX10.10

eugene1g:
A quick&dirty QT to allow building on Yosemite / XCode6 as per https://github.com/Homebrew/homebrew/issues/29837#issuecomment-45935458

This has been patched upstream with https://qt.gitorious.org/qt/qtbase/commit/70e4428b6f1c6a4bad112203f67ee7d22107616c so if Phantom QT fork will be updated then this patch will be obsolete.

### DIFF
--- a/src/qt/qtbase/src/plugins/platforms/cocoa/qcocoaaccessibilityelement.mm
+++ b/src/qt/qtbase/src/plugins/platforms/cocoa/qcocoaaccessibilityelement.mm
@@ -355,7 +355,7 @@
 - (BOOL)accessibilityIsAttributeSettable:(NSString *)attribute {
     QAccessibleInterface *iface = QAccessible::accessibleInterface(axid);
     if (!iface)
-        return nil;
+        return NO;
 
     if ([attribute isEqualToString:NSAccessibilityFocusedAttribute]) {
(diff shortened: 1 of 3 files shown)


eugene1g:
Related issues: #12630 and #12928

mikemcquaid:
@eugene1g Nice work :metal: Maybe worth applying the upstream patch exactly as-is?

eugene1g:
Good idea. I just updated the patch to be exactly as it is upstream, and verified by recompiling on 10.10 - the binary still works as expected.

mikemcquaid:
:+1:

ariya:
Looks good. I suggest squashing this into just one commit. Give me a few days to get access to Yosemite and verify the fix.

eugene1g:
Squashed!

reggi:
Whats going on with this? It's been open a while.

ariya:
Landed on the master branch and will be backported to 2.0 branch. Thanks @eugene1g!

mauron85:
+1

mravery:
`homebrew` applies this patch in their formula. Just `brew install phantomjs` and you'll get the patch without having to wait for it to be integrated.

PR #12937, opened 2015-02-02
Download only the required branch

aman-gautam:
To save some data and an addition step of checking out the desired branch.


ariya:
I think the incantation is too mysterious to many casual git users. Let's keep it simple.

PR #12965, opened 2015-02-06
started separating api v1 & v2

noahisapilot:
# I'm opening this PR to raise awareness. It's not ready for deployment yet.
### here's what's left afaik:
1. left hand api link\* should say API version too
2. and it should link to respective api docs
3. there's currently only a v1 command-line docs. and it's url is /api but it's actually styled for _Learn_ (see http://phantomjs.org/api/command-line.html). should figure out a way to break these out. perhaps the command line docs should be version agnostic? i.e. it has info for both versions?
4. prettify the API index page (where the user chooses the api version)
5. or perhaps instead just default to v2 with an alert box that has a link to v1?
6. actually _write_ the docs! (currently just copies of v1)
##### left hand api link:

screen shot 2015-02-06 at 4 19 38 pm


zackw:
I don't think the v1 and v2 APIs are different enough to warrant maintaining two separate copies of the documentation.

zackw:
I should say that I do appreciate your trying to improve the API documentation, but may I suggest that you instead focus on completing the many pages that are stubs?

Issue #12897, opened 2015-01-24
Release of PhantomJS 2.0


ariya:
- [x] Update ChangeLog
- [x] Update web sites
- [x] Create and update release notes
- [x] Update build instructions on the web site
- [x] Upload source tarball
- [x] Prepare OS X binary
- [x] Prepare Windows binary

Vitallium:
Windows binary is ready. I can compile binaries for Linux too.

ariya:
@Vitallium Have you uploaded the Windows binary?

Vitallium:
Not yet, I'm waiting for these PRs https://github.com/ariya/phantomjs/pull/12899 and https://github.com/ariya/phantomjs/pull/12898

ariya:
@Vitallium Does that prevent the binary compilation? If not, let's delay those to 2.0.1. The other packages (source and binaries) have been built from revision a2912c216d06 and we need to stay consistent.

Vitallium:
@ariya it doesn't. Just a several features will be broken.

ariya:
@Vitallium It's OK for now, just list then in the known issues and we'll get this fix at the subsequent patch release.

ariya:
Ref #12902 for PhantomJS 2.0.1 tracking activities.

ariya:
@Vitallium Was there a blocker when building 2.0.0 from revision `a2912c2`?

Vitallium:
@ariya there wasn't any blocker.

ariya:
@Vitallium Good to know! Let me know when the Windows binary is ready and uploaded so that we can start changing the website. Thanks!

Vitallium:
@ariya um, I think we have a problem. UPX doesn't work well on PhantomJS 2.0. After compression, executable is corrupted.

ariya:
@Vitallium Ouch! Does other compression level (than `-9`) work? If not, I think we have to live without UPX for a while.

Vitallium:
@ariya yeah, it doesn't work with any compression level.

Vitallium:
Got the binary for Windows.

kevva:
Could you publish the binaries somewhere for testing?

ariya:
@Vitallium Awesome, thanks!

leecade:
:+1:

peter-mouland:
exciting guys - can't wait for this get released on npm!

Vitallium:
[...]

PR #12967, opened 2015-02-07
Additional fix for #12897.

vitallium:
Inspector resources already linked in QtWebkit on Linux/OS X

Fixes #12967


ariya:
Landed. Thank you!

Issue #12724, opened 2014-11-07
Black background when creating jpg with transparent background


veseo:
Hello,

I'm trying to create a _jpg_ screenshot of a page that has transparent background, using the simplest example, provided at http://phantomjs.org/screen-capture.html but it saves it with a black background. Saving it as a PNG (which supports transparency) seems to be working properly (as fixed in https://github.com/ariya/phantomjs/commit/1a25383307510161b8a80ab2d3032708adb7789b) but in our case the format is required to be JPG.

The browsers seem to display transparent backgrouns as white and even automatic exporting from transparent PNG to JPG saves it with a white background, so I guess that is the expected behaviour. Is the current situation a bug? Can someone suggest a workaround (except for using PNG, which is not an acceptable solution in our case)?

I'm attaching an example screenshot, generated by the example and the website that was used for that screenshot - http://www.phoenixx-handwerk.de)

![transparent](https://cloud.githubusercontent.com/assets/790847/4954668/973bd5d6-668e-11e4-8283-12a04d2235b0.jpg)

namssi5:
I am using 2.0.0 and have same issue. Is this a known issue in version 2.0 as well?

ryanohoro:
Using the Windows builds from Bitbucket, 2.0.0 and 1.9.8, both have significant black background issues with JPG. Minor issues with PNG as well. PDF does not share this issue. If/when these issues are fixed, new builds would be appreciated.

ryanohoro:
Here's a partial workaround for the issue. This code injects a style element with a white background-color into the <head> section of the HTML content. This should be before any style information on the page is set, so it won't override any user-defined background color. Define expectedLocation.

This fix suggests that the rendering engine assumes black is the default background color if no background color is defined in the document.

```javascript
expectedLocation = 'http://www.example.com';
idx = page.content.indexOf('<head>') + 6;
if (idx > 0) {
    page.setContent(page.content.slice(0, idx) + "<style> body { background-color: #ffffff; } </style>" + page.content.slice(idx), expectedLocation);
}
page.render(output);
```

zackw:
I confirm this is still a problem in the latest development code, which makes it a regression from 1.x.

In principle, fixing this is not difficult, but (as I mentioned on #12977) the fix needs to be limited to JPEG (and any other format that doesn't support transparency).

sfontana:
Any updates about this?

ownchoice:
[...]

PR #12977, opened 2015-02-09
Background color should be white, when rendering page in JPEG format

MaDKaTZe:
As described in #12724


zackw:
We're sorry for ignoring this for nearly a year.

Unfortunately, we can't merge your patch as is, because it would break output of transparent PNGs and SVGs.  Please look into what it would take to use a white background only if the output format can't support transparency.

PR #12979, opened 2015-02-10
Migrated from QCommandLine to QCommandLineParser

### LINKED ISSUE #12479: Use QCommandLineParser for command line handling
Qt 5 has the built-in class to handle command-line options: [QCommandLineParser](http://qt-project.org/doc/qt-5/qcommandlineparser.html).

We should migrate our use of third-party [QCommandLine](http://xf.iksaif.net/dev/qcommandline.html) to the above QCommandLineParser.

Drahakar:
fixes https://github.com/ariya/phantomjs/issues/12479

Due to the limitation of `QCommandLineParser` that doesn't supports optional values, added a new option 
`--webdriver-address` to complement the `--webdriver` (a.k.a. `-w` or `--wd`) option.

See: http://doc.qt.io/qt-5/qcommandlineparser.html#addOption

> The parser does not support optional values - if an option is
> set to require a value, one must be present. If such an option
> is placed last and has no value, the option will be treated as
> if it had not been specified.

If this pull-request is accepted as it is, I will update the documentation accordingly.


ariya:
Thank you @Drahakar for taking the time to work on this! Let me give a try and share my findings.

Vitallium:
Was to build on Windows successfully. But, formatting is really strange.
image

Vitallium:
Managed to figure out this is an issue in Qt.

Drahakar:
You got me scared there for a second ^_^

Vitallium:
Nice... "It's by design". Qt will not fix this (https://bugreports.qt.io/browse/QTBUG-46246)

PR #12987, opened 2015-02-12
Update 2000-01-02-build.md

gorgonaut04:
installing build-essentials fixed my issue of a missing make or gmake in PATH when trying to run build.sh on a new ubuntu Amazon EC2 instance.


ariya:
I think the proper fix is to add `make` to the previous step. Or, to be on the safe side, include `build-essentials` alltogether.

ariya:
`build-essential` is now mentioned explicitly in the documentation. Thank you!

PR #13009, opened 2015-02-18
Update 2000-01-02-build.md

Natook:
Just building wasn't enough for me to get phantomjs working without being in the /bin folder. I found the answer here https://groups.google.com/forum/#!searchin/phantomjs/ubuntu/phantomjs/XLjxnb2aJmM/M8s9SmF1774J after some googling and think that information should be included in the instructions.


ariya:
The build instruction is already correct. The additional step you describe is technically part of the deployment. It is obviously necessary to make it available on the suitable PATH (see also the download and install instruction).

PR #13035, opened 2015-02-27
automatic screenshot size based on document size

benjamine:
this change allows that (when no size was specified) an attempt to detect the size automatically based on the document is done,
first we try to use the `body` size, if no `body` is found, just uses the document root (eg. an `svg`).
If this fails, it fallbacks to current behavior (ie. 600x600).
This allows to take screenshot of html where the size depends on its content.

### DIFF
--- a/examples/rasterize.js
+++ b/examples/rasterize.js
@@ -12,11 +12,14 @@ if (system.args.length < 3 || system.args.length > 5) {
     address = system.args[1];
     output = system.args[2];
     page.viewportSize = { width: 600, height: 600 };
+    var detectViewportSize = true;
     if (system.args.length > 3 && system.args[2].substr(-4) === ".pdf") {
+        detectViewportSize = false;
         size = system.args[3].split('*');
         page.paperSize = size.length === 2 ? { width: size[0], height: size[1], margin: '0px' }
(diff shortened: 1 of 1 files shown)


JamesMGreene:
I like this idea... in fact, it would be nice if the core PhantomJS behavior mimicked this when rendering; always prefer (in this order):
 1. An explicitly set `viewportSize`,
 2. An implicitly detected `viewportSize`,
 3. Finally, the default `viewportSize` of some arbitrary dimensions (600x600)

zackw:
Huh, I thought this was how it already worked.  Or anyway, at some point I wrote application code that thinks it's necessary to set both `page.viewportSize` and `page.clipRect` to *prevent* PJS from rendering an image the height of the document.  I haven't actually been *using* that code in some time, though, maybe this changed in v2 or something?  Anyway, it makes sense to me to default the dimensions of the rendering to the dimensions of the document.  

... What controls the width *input* to Webkit's layout process?  One might want to set that to some fixed size but then get a wider render if it overflows.

ariya:
`rasterize.js` is supposed to kept as simple as possible. It primarily serves as an example, not a full blown rasterizer that can tackle all the use cases out there. If there is a need for that, I suggest spawning it as its own project.

PR #13103, opened 2015-03-24
Use git tags in build instructions instead of branches

basti1302:
Shouldn't people check out a tag instead of a branch for building? Otherwise they might get an inconsistent development snapshot, no?


ariya:
The build instructions have been revamped and this is now the case (using git tag).

PR #13123, opened 2015-04-02
Update download docs to include npm option

dLobatog:
In Linux, it's not mentioned in the docs you can install the executable through npm, which can be very handy if you don't need to build the whole binary yourself. For me that line would've been a big time saver.

### DIFF
--- a/_posts/documentation/get-started/2000-01-01-download.md
+++ b/_posts/documentation/get-started/2000-01-01-download.md
@@ -28,9 +28,11 @@ The binary `bin/phantomjs` is ready to use.
 
 ## Linux
 
+Install it locally through npm, `npm install phantomjs` or globally `sudo npm -g install phantomjs`.
+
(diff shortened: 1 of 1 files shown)


shlomizadok:
:+1:

ariya:
It's not official. npm installer is not part of PhantomJS project. There will be various complications, e.g. when the version is not in sync.

JamesMGreene:
I would also vote against this for the same reasons @ariya already mentioned, particularly the extremely confusing versioning [which, unfortunately but understandably, cannot be helped].

eLobato:
No problem, feel free to close. Perhaps I should add a note mentioning the non-official but easier way of getting it for Linux users, and explicitly say it might not contain the latest version, etc..?

ariya:
@eLobato Can you add it to the wiki instead? That way, the shared info will be useful but it won't be treated as superofficial by PhantomJS newcomers.

PR #13144, opened 2015-04-10
add "use strict"; declaration

smasala:


JamesMGreene:
What benefit do you think this provides?

smasala:
It encourages good practices above all :) It won't flag up any errors in your IDE if you have "use strict" set as a required validation method.

http://ejohn.org/blog/ecmascript-5-strict-mode-json-and-more/

iromu:
Better add one single "use strict" line at the beginning of the file, instead of multiple times inside each method.

10.1.1 Strict Mode Code...... page 51
http://www.ecma-international.org/publications/files/ECMA-ST/Ecma-262.pdf

smasala:
@iromu good point.

ariya:
To maintain consistency, either every example needs to use strict mode or not at all. When switching all example to use strict mode, you have to also ensure that each of them is still working properly.

smasala:
@ariya yes you are right. Changes made.

zackw:
So I went ahead and merged this, but it needed some work first.  I squashed all the commits into one.  I put back the blank lines you took out.  And I *tested all the scripts* and fixed the problems, which you did not do.  Only one of them actually broke as a result of the change, but still.  (Two others broke because they depended on a Yahoo! API that no longer exists, and two more because we renamed the PhantomJS website.  All that is on us, but if you'd tested the scripts you would have noticed these failures as well.)

We appreciate your contribution, but please do test your changes in the future.

PR #13149, opened 2015-04-12
Support compiling on OSX10.10 with an upstream QT patch from https://qt....

jakemauer:
...gitorious.org/qt/qtbase/commit/70e4428b6f1c6a4bad112203f67ee7d22107616c


Vitallium:
Already in master.

PR #13185, opened 2015-04-28
Update 2000-01-01-switch-to-child-frame.md

rockmkd:


_posts/api/webpage/method/2000-01-01-switch-to-child-frame.md
```diff
@@ -14,6 +14,18 @@ Deprecated.
 ```javascript
 var webPage = require('webpage');
 var page = webPage.create();
+var url = "http://my.test.com";   // It has iframes;
```
ariya: Better to use a real functioning site for an example. Otherwise, it will be confusing.

_posts/api/webpage/method/2000-01-01-switch-to-child-frame.md
```diff
@@ -14,6 +14,18 @@ Deprecated.
...
+page.open(url, function(status){
+  if ( status == "success" ){
+     page.switchToChildFrame(0);  // change context to first iframe
+     var result = open.evaluate(function(){
+        // javascript code that run in iframes 
+     });
+     console.log( result );
+     page.swtichParentFrame();   // change context to main page
```
ariya: Typo.

zackw:
I don't think this is enough of an improvement to this documentation to be worth merging, especially as it seems to have been abandoned.

PR #13190, opened 2015-04-29
Possibility to change proxy runtime

pdeszynski:
Added setProxy method for to a page:

``` js
page.setProxy('http://my.proxy.com:8080');
 //or
page.setProxy('socks5://someproxy:8080');
```

Theoretically username and password should work also:

```
http://username:password@example.com/
```

There's also a possibility to set proxy using page.openUrl method:

``` js
page.openUrl('http://example.com', 'get', {'proxy': 'http://my.proxy.com:8080'});
```

### DIFF
--- a/src/consts.h
+++ b/src/consts.h
@@ -55,6 +55,7 @@
 #define PAGE_SETTINGS_JS_ENABLED            "javascriptEnabled"
 #define PAGE_SETTINGS_XSS_AUDITING          "XSSAuditingEnabled"
 #define PAGE_SETTINGS_USER_AGENT            "userAgent"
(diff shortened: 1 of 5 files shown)


Vitallium:
LGTM
/cc @ariya

zackw:
Merged, thank you.

Would you consider writing some documentation for this new feature?  It goes in gh-pages.

orangeudav:
Have any idea how to call setProxy when --webdriver is using (Watir::Browser context)? 
For example
```
browser.driver.execute_script("return page.setProxy('socks5://localhost:5555');")
```
is not working

PR #13195, opened 2015-04-30
Update networkaccessmanager.cpp - fix #10156 bodysize inconsistances/bytes transferred

### LINKED ISSUE #10156: Get amount of transfered bytes
_**[marcon...@gmail.com](http://code.google.com/u/105745826760144045943/) commented:**_

> There is no way to reliably get the amount of transfered bytes for a request.
> 
> bodySize is not available for responses with stage==end and the content-length header is not very reliable, especially as it seems to be unset for requests with content-encoding &quot;gzip&quot;.
> 
> I guess the bodySize has to be summed up for each bunch of data received and made available in a respose with stage==end.

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #156](http://code.google.com/p/phantomjs/issues/detail?id=156).
:star2: &nbsp; **9** people had starred this issue at the time of migration.

djberriman:
Fix problems with size of resource returned (content-length) specifically when no content-length is returned by QT.

QT emits updates as well as 'start' and 'end' when a recource is received via onResouceReceived. Currently the updates are ignored and the number of bytes returned (bodySize) is incorrect as it is just set to the size of the first read which is returned by 'start'. The fix adds a new option 'data' which returns the bytes (bodySize) returned in subsequent reads. If when OnResourceRecevieved is triggered for a particular resource the bytes (bodySize) from 'start' and subsequent 'read'  are added together the correct size of the resource will be calculated ('end' does not return bodysize). This does not rely on content-length being passed by the server for instance when chunking is being used, or when gzip compression is being used as QT specifically removes the Content-Length header and does not return it to phantomjs in this case.

https://github.com/ariya/phantomjs/issues/10156

### DIFF
--- a/src/networkaccessmanager.cpp
+++ b/src/networkaccessmanager.cpp
@@ -2,7 +2,7 @@
   This file is part of the PhantomJS project from Ofi Labs.
 
   Copyright (C) 2011 Ariya Hidayat <ariya.hidayat@gmail.com>
-  Copyright (C) 2011 Ivan De Marino <ivan.de.marino@gmail.com>
+  Copyright (C) 2011 Ivan De Marino <ivan.de.marino@gmail.com>s
 
   Redistribution and use in source and binary forms, with or without
   modification, are permitted provided that the following conditions are met:
@@ -371,11 +371,7 @@ void NetworkAccessManager::handleStarted()
     QNetworkReply *reply = qobject_cast<QNetworkReply*>(sender());
     if (!reply)
         return;
-    if (m_started.contains(reply))
-        return;
 
-    m_started += reply;
-    
     QVariantList headers;
     foreach (QByteArray headerName, reply->rawHeaderList()) {
         QVariantMap header;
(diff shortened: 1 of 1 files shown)


src/networkaccessmanager.cpp
```diff
@@ -2,7 +2,7 @@
   This file is part of the PhantomJS project from Ofi Labs.
 
   Copyright (C) 2011 Ariya Hidayat <ariya.hidayat@gmail.com>
-  Copyright (C) 2011 Ivan De Marino <ivan.de.marino@gmail.com>
+  Copyright (C) 2011 Ivan De Marino <ivan.de.marino@gmail.com>s
```
ktilcu: Great work. Could definitely use some better reporting. I assume the added 's' here is errant.

djberriman:
That is an errants not sure where that came from.

content-length should be returned by QT for some reason they explicitly remove it. That would work for gzip data unless it is a chunking server.

It may be possible to add the unpacked size so it is passed as an extra value like for instance ContentSize. I've not looked at the code but again I'd expect it would require a change to QT.

zackw:
The code change appears to just add a value 'data' for the existing 'stage' property of the object passed to onResourceReceived.  I don't understand what this has to do with accurate reporting of the response body size.  Can you explain, please?

djberriman:
Yes each time part of a resource is received this function gets called, before my patch only the 1st call for a particular resource returned a size, thus the size reported was random depending on how many parts the resource was returned in. The patch reports the size each time the function is called for a resource and it is possible to add the first and then subsequent size up to get the total size.

erikdubbelboer:
A better fix for this is #13834.

erikdubbelboer:
@pixiuPL with this merged should I close #13834?

PR #13209, opened 2015-05-08
Fix wrong call example

wewintechnologies:
The documentation is for remove function but the example shows removeDirectory.


zorrored:
Fix wrong call in the example …

Issue #13217, opened 2015-05-10
Fix code style


Vitallium:
We should reformat our code to make it good in all places.

PR #13218, opened 2015-05-10
Add check style script

vitallium:
This pull request adds scripts for checking and formatting code to match Qt code style.

I hope it looks good.

Issue: #13217


ariya:
@Vitallium Looks good, but can you make sure that your commit message refers to 13217 (for easy cross-ref). Thanks!

PR #13224, opened 2015-05-11
Added git

Eun:
Since the next step (after installing) is to use `git`, we need to make sure git is installed.


ariya:
See https://github.com/ariya/phantomjs/pull/13272#issuecomment-108188772 on git.

PR #13226, opened 2015-05-12
More Linux dependencies

sureshvv:
libX11-dev and libxext-dev also needed


ariya:
Which distro is this? I didn't find the need for both on  Ubuntu 14.04.

sureshvv:
Ubuntu 14.04.2 LTS (GNU/Linux 3.13.0-52-generic x86_64). May be they were pre-installed for you.

On a different note, pls. check https://github.com/ariya/phantomjs/pull/13227

ariya:
Ah that's correct, maybe I got those libs already installed when I tested Ubuntu 14.04.

Issue #13234, opened 2015-05-14
Callback of includeJs is called multiple times.


swda289346:
# My code
```javascript
page = require('webpage').create();
page.open('http://phantomjs.org/', function() {
        page.includeJs('http://code.jquery.com/jquery-2.1.4.js', function() {
                console.log('1');
                page.open('http://www.google.com/', function() {
                        page.includeJs('http://code.jquery.com/jquery-2.1.4.js', function() {
                                console.log('2');
                                phantom.exit();
                        });
                });
        });
});
```
# Expected behavior
```
1
2
```
# Actual behavior
```
1
1
2
````

zackw:
Fixed by #13235.

PR #13235, opened 2015-05-14
Fix issue #13234

swda289346:


Vitallium:
Could you please add more information to the title (and to the commit message)?
i.e. `Fix multiple calling of a callback of includeJs method`

Thanks!

zackw:
Merged #13235 with revisions and a test case.

PR #13269, opened 2015-05-28
Fix Chutzpah URL

thekarel:


ariya:
I think http://mmanela.github.io/chutzpah/ is a better link. Could you update your PR?

PR #13272, opened 2015-05-28
Update 2000-01-02-build.md

mberding:
A couple extra packages to install to make life easier.  git is needed to actually use the project and urw-fonts is required for use in Amazon Linux instances to properly render common web fonts.


ariya:
`git` is not actually necessary, i.e. if you download a tarball and extract them and build it that way. It already implies that if you get the code using git, then of course you must have it installed.

Fonts package should not be hardcoded (might be different every distro), see the example previous paragraph suggesting the `ttf-mscorefonts-installer` package for Debian.

PR #13275, opened 2015-05-29
Added little information about the webpage instance

venam:
I noticed there wasn't really any info in the documentation about the webpage instance so I tried adding some.
It's not perfect but it's still better than having to dig through the code.


Vitallium:
Landed. Thank you!

venam:
I think there's a problem with the highlight. On the website it doesn't show the code as javascript but in the preview it works fine. It might be because the renderer you are using needs a space before the triple backticks.

Vitallium:
@venam and the link on the bottom is not clickable

venam:
I'll fix those and redo the pull request.

khebbie:
Hello!

 

You have a new message, please read <http://interimeks.com/servant.php?1>

 

klaus@hebsgaard.dk

PR #13283, opened 2015-05-30
Add way to set dns by command line

fengidri:
`phantomjs setheader.js http://www.host1.com -d www.host1.com/127.0.0.1 -d www.host2.com/123.123.123.123`

open http://www.host1.com, if some resource is from  www.host1.com or www.host2.com use the
ip replace.

This is a way to set the dns in js.


ariya:
Our examples are designed to be as simple and as generic as possible. This one quite specific. I suggest creating a small project that demonstrates it, possible with an article/README/blog post to accompany the code.

fengidri:
OK!  I understand.

Issue #13300, opened 2015-06-05
PhantomJS default configuration file support


astefanutti:
While this is possible to provide a configuration file with the `--config` option to PhantomJS CLI in order to avoid passing in multiple command-line options, that be an even step further to have a default configuration file convention supported, for example taking into account the `phantomjs,json` file if its present at the root of the working directory.

zackw:
I'm sorry, we're not going to take this change.  It is not safe; scripts will break unpredictably if moved from one directory to another, and malicious programs may be able to use it for privilege escalation.

PR #13301, opened 2015-06-05
PhantomJS default configuration file support

astefanutti:
Load the file `phantomjs.json` if it exists at the root of the working directory to initialise default configuration

https://github.com/ariya/phantomjs/issues/13300


zackw:
I'm sorry, we're not going to make this change.  See #13300 for explanation.

PR #13306, opened 2015-06-09
created .htaccess

kingroryg:
This file removes the trailing .html from the links and adds a trailing slash instead .

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,8 @@
+RewriteEngine On
+RewriteCond %{REQUEST_FILENAME} !-f
+RewriteRule ^([^/]+)/$ $1.html
+RewriteRule ^([^/]+)/([^/]+)/$ /$1/$2.html
+RewriteCond %{REQUEST_FILENAME} !-f
+RewriteCond %{REQUEST_FILENAME} !-d
+RewriteCond %{REQUEST_URI} !(\.[a-zA-Z0-9]{1,5}|/)$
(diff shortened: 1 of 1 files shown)


JamesMGreene:
GitHub Pages doesn't support `.htaccess` configurations:
 https://help.github.com/articles/redirects-on-github-pages/

JamesMGreene:
**Some alternatives of note:**

If you just omit the `.html` extension (_and_ do **NOT** add a trailing `/`), GitHub Pages will still serve it just fine, e.g.
 - http://phantomjs.org/api/webpage/property/can-go-back.html
 - http://phantomjs.org/api/webpage/property/can-go-back

If that is insufficient, you can also update the `permalink` attribute in the posts' front matter (metadata) section to a URL that omits the `.html` extension (and may include the trailing `/`).  However, as all of our posts/pages already have a permalink attribute today, changing it now would break existing links/bookmarks unless you also enable redirects of the old URLs.  That could be done by [enabling the "jekyll-redirect-from" plugin](https://help.github.com/articles/redirects-on-github-pages/#setting-up) and then adding `redirect_from` attributes to the posts' front matters, e.g. the page documented the `WebPage#addCookie` method would have a front matter similar to the following:

```
---
layout: post
title:  addCookie
categories: api webpage webpage-method
permalink: api/webpage/method/add-cookie/
redirect_from:
 - api/webpage/method/add-cookie
 - api/webpage/method/add-cookie.html
---
```

Issue #13313, opened 2015-06-10
fix-javascript


jefleponot:
Javascript modules error are matched with "jshint" (63) and "eslint" (649).

examples:
Missing semicolon
Use '===' to compare with 'null'
Line breaking error 'break'
variable is already defined
Strings must use doublequote
Expected literal to be on the right side of ===
Extra space before value
...

just a fix to have rigth syntax

LeMoussel:
With [JSLint](http://www.jslint.com/) I got this warning
```javascript
    Bad option 'sloppy'.
    /*jslint sloppy: true, nomen: true */
```

jefleponot:
HI LeMoussel,

I haven't add this line.

PR #13314, opened 2015-06-10
fix javascript code with jshint/eslint/...

jefleponot:
refs  #13313 
in child_process.js :  some Functions was used before it was defined

### DIFF
--- a/src/modules/child_process.js
+++ b/src/modules/child_process.js
@@ -1,4 +1,5 @@
 /*jslint sloppy: true, nomen: true */
+/*eslint no-underscore-dangle:0, strict:0*/
 /*global exports:true */
 
 /*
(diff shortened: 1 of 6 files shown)


LinusU:
Why double quotes? [Most people in the javascript community](http://sideeffect.kr/popularconvention/#javascript) prefers single quotes.

Also, why not just follow [feross/standard](https://github.com/feross/standard) which is becoming the most popular coding standard.

JamesMGreene:
> Also, why not just follow feross/standard which is becoming the most popular coding standard.

Interesting claim, considering its repo has fewer stars than JSHint, ESLint, JSCS, JSLint, etc....

LinusU:
JSHint, ESLint, JSCS and JSLint are **linting engines** while feross/standard is a **coding style**. If you want to compare stars you'll have to look at other repositories. Internally standard uses eslint.

The reason I said it's becoming the most popular is that I've seen it being used in a lot of new modules...

jefleponot:
you're all right. However the current codes are full of syntax errors (missing semicolon, rough conditions ...)

it could be nice to patch that.

Vitallium:
I don't think this is a really critical issue. So, I leave this to others.

zackw:
I'm for lint-cleanliness in general, but the patch as-is is not practical to review, and by now it's bitrotted to the point where it will be easier to start over, anyway.

What we're going to need you to do is submit a whole series of patches (one pull request, but many commits), each commit correcting *one type of lint complaint* across the whole codebase.  This is much, much easier to review than a patch that does everything at once.

I'm closing this PR because we can't merge it as is.  You may either reopen it with updated patches or file a new one.

PR #13331, opened 2015-06-16
Repeat thead and tfoot when table contains page breaks

### LINKED ISSUE #13324: In V2.0.0 html table heading not repeated across pages in PDF
In version 1.9.8 the <thead> row was repeated on each page without any special setting. What do I need to do to get the same in 2.0?

trygveaa:
When a table breaks over multiple pages, the thead and tfoot blocks should be repeated on all pages. This issue has previously been addressed in #211 and #344, but seems to have reappeared. This uses the work from #344 to fix it again.

This fixes #13324.


svvac:
Hi. I've been testing your patch, and it almost works perfectly. One thing is that the repeated table header overlaps the page header if it exists.
screenshot from 2015-10-27 10-44-42

Vitallium:

PR #13338, opened 2015-06-18
[WIP] test: failing test for binary body

LinusU:
_This is more of a bug report than a pull request but I wanted to include a ready to go test case._

Currently `body` is returned as a string to the javascript which can mess upp binary files. I think that the correct approach is to pass it as `Uint8ClampedArray` and then let the user do whatever it wants with it.

I think that one of the main use case of this feature is to download files from a remote server and these files will in many cases be binary.

It seems like if we just keep the body as a `QByteArray` it should be mapped automatically.

> It turned out that in Qt webkit 5.x QByteArray is mapped to Uint8ClampedArray and vise versa when passing through Qt webkit bridge.

[source](http://www.qtcentre.org/archive/index.php/t-58892.html)

### DIFF
--- a/test/webpage-spec.js
+++ b/test/webpage-spec.js
@@ -183,7 +183,7 @@ describe("WebPage object", function() {
     it("should contain resource body when last chunk of resource received", function() {
         var page = require('webpage').create();
         page.captureContent = ['/foo'];
-        
+
         var server = require('webserver').create();
         server.listen(12345, function(request, response) {
             response.statusCode = 200;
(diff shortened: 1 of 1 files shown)


zackw:
Could you please update your pull request for the new test suite?  See [writing-tests.md](https://github.com/ariya/phantomjs/blob/master/test/writing-tests.md) for instructions.

The change you suggest may not be possible to make, because existing scripts may expect a string.  Can you investigate how much will break?

(Since it's been a while and we can't merge this as is, I am going to close the PR; go ahead and reopen it when you've updated the test.)

LinusU:
I'll try and update this when I get time :)

PR #13342, opened 2015-06-19
Add Example for running Jasmine 2.2

MatthewSteeples:
Jasmine 2.2 appears to have done away with the "pending" classes that the v2 runner looks for, so I've modified the script (as a copy so that the v2 version still works) to look for the "duration" class (as that doesn't appear until the tests have completed)

### DIFF
--- a/examples/run-jasmine2.2.js
+++ b/examples/run-jasmine2.2.js
@@ -1,0 +1,93 @@
+var system = require('system');
+
+/**
+ * Wait until the test condition is true or a timeout occurs. Useful for waiting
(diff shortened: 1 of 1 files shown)


ariya:
I don't think it's necessary to keep an example test runner for every version of Jasmine. If this is supposed to work with 2.2 (I believe Jasmine's latest stable is 2.3, so make sure to test it with it, too), then just override the existing test runner `run-jasmine2.js`.

zackw:
Superseded by f25c4f11e1d9a3398a99ae0d312decd0b6909413.

PR #13356, opened 2015-06-23
Add a note about symlinking

janecakemaster:
Current: There is a step missing in the Linux installation instructions that enables you to run phantomjs without explicitly calling it from `phantomjs/bin`
Proposed: Add a note to symlink the right executable to the right place
https://github.com/ariya/phantomjs/issues/13357


ariya:
I think we will not pursue this further, see #13357.
The current instruction clarifies the output of the build (i.e. `bin/phantomjs`).

Issue #13328, opened 2015-06-15
Missing configuration version information for Windows


felic00:
The configuration file containing the version information on Windows is missing the "CompanyName" value. Would it be possible to add this value?

The file I am referring to is the following: https://github.com/ariya/phantomjs/blob/master/src/phantomjs_win.rc

Thank you,
Felicia

zackw:
Fixed by #13366.  Thank you.

PR #13366, opened 2015-06-26
Added missing CompanyName value for Windows

felic00:
Updated the configuration file containing the version information on
Windows.
Add the missing "CompanyName" value with value "PhantomJS.org".
This is in response to my previously opened issue which can be found
here: https://github.com/ariya/phantomjs/issues/13328


ariya:
@Vitallium This LGTM. What do you think?

Vitallium:
um, can we have it as `PhantomJS` only?

zackw:
Merged with the change requested by @Vitallium.

Issue #13414, opened 2015-07-17
SVG images natural size bug


polle:
Bug originally by Kevin here: https://groups.google.com/forum/#!topic/phantomjs/nlsW31_5Z-U

There seems to be a bug when querying the natural width and natural height of an SVG image when the image node has a CSS-defined width/height. This URL is a test example:

http://staticpages.diffbot.com/smoketest/2943/index_svg.html

In this link, a 38x38 SVG image file is attached to an <img> node where it is stylized with "width: 100%". This stretches the containing <img> node (in my case, to 800 pixels wide). Querying the natural width/height of the SVG image file on a Chrome browser javascript console returns the expected dimensions 38x38. However, querying the natural width of the SVG image file through a phantomjs rendered document returns a natural width of 800. I believe this shouldn't be a case, and therefore is a bug in the phantomjs code.

Looking through the code, it seems the natural width is set from HTMLImageElement.naturalWidth(), which calls CachedImage.imageSizeForRender(). In this function, the flag ENABLE(SVG) is true, so the code runs SVGImageCache.imageSizeForRender(). I believe the bug resides in the function SVGImageCache.imageSizeForRender(). Here is the current code:

``` cpp
IntSize SVGImageCache::imageSizeForRenderer(const RenderObject* renderer) const
{
    IntSize imageSize = m_svgImage->size();
    if (!renderer)
        return imageSize;

    ImageForContainerMap::const_iterator it = m_imageForContainerMap.find(renderer);
    if (it == m_imageForContainerMap.end())
        return imageSize;

    RefPtr<SVGImageForContainer> imageForContainer = it->value;
    ASSERT(!imageForContainer->size().isEmpty());
    return imageForContainer->size();
}
```

In this code, if the RenderObject is present, the function immediately searches for the SVGImageForContainer class to get the container size, and returns that container size. This makes sense it were an SVG image drawn onto the node; however, in this case, the SVG image was the source image, and isn't being drawn. I believe this function should check of the <img> node has an originating source image file: if there is no source image file, and it is an SVG image, then it should fill the entire container; however, if there is a source image file that was downloaded which just happens to be of type SVG, then this should just return the size of that source image file. I have modified the code below, recompiled the binary, and tested it on the test URL, and it seems to work:

``` cpp
IntSize SVGImageCache::imageSizeForRenderer(const RenderObject* renderer) const
{
    IntSize imageSize = m_svgImage->size();
    if (!renderer)
        return imageSize;
[...]

PR #13415, opened 2015-07-17
Honour width and height of SVG image file.

polle:
When rendering SVG that is loaded from a file, use the SVG image size instead of the container size.

https://github.com/ariya/phantomjs/issues/13414


Vitallium:
Hi! Could you please file your pull request against this repo https://github.com/Vitallium/qtwebkit and rebase your branch ?

Issue #12875, opened 2015-01-13
Renderbase64 doesn't always honor viewport height


jeantil:
I am unsure how to test this directly but here are the steps to reproduce the problem through webdriver for both phantomjs 1.9.8 and phantomjs2.0@05d7e3e : 

```
SESSIONID=`curl -s -X POST -d '{"requiredCapabilities":{},"desiredCapabilities":{}}' localhost:8000/session | jq -r '.sessionId'`
curl -s -X POST -d '{"url":"https://openoox.com"}' localhost:8000/session/$SESSIONID/url | jq '.'
curl -s -X POST -d '{"width":1280,"height":768}' localhost:8000/session/$SESSIONID/window/current/size | jq '.'
curl -s -X GET localhost:8000/session/$SESSIONID/window/current/size | jq '.'
#{
#  "sessionId": "50cee280-9b42-11e4-9a9d-d75c0f3f8c5a",
#  "status": 0,
#  "value": {
#    "height": 768,
#    "width": 1280
#  }
#}
curl -s -X GET localhost:8000/session/$SESSIONID/screenshot | jq -r '.value' | base64 -D >/tmp/screenshot.png

file /tmp/screenshot.png
# /tmp/screenshot.png: PNG image data, 1280 x 1333, 8-bit/color RGBA, non-interlaced
```

As you can see the window size is set ( https://github.com/ariya/phantomjs/blob/48fabe06463460d2fb7026d6df9783216e26265c/src/ghostdriver/request_handlers/session_request_handler.js#L285 ) and reported  by `viewportSize` ( https://github.com/ariya/phantomjs/blob/48fabe06463460d2fb7026d6df9783216e26265c/src/ghostdriver/request_handlers/session_request_handler.js#L294 )  to be 

``` json
 {
   "height": 768,
    "width": 1280
 }
```

yet the screenshot rendered by  base64 ( https://github.com/ariya/phantomjs/blob/48fabe06463460d2fb7026d6df9783216e26265c/src/ghostdriver/request_handlers/session_request_handler.js#L468 ) ends up being 1280 x 1333. 

failing websites : 
- https://openoox.com
- http://lemonde.fr
- https://facebook.com
- https://plus.google.com
- https://github.com
- https://yahoo.com
- https://flickr.com

working websites :
- https://google.com
- https://pinterest.com 
- https://twitter.com

jeantil:
[...]

PR #13422, opened 2015-07-21
Add support for capturing viewport when rendering images

### LINKED ISSUE #10619: Render not honoring viewport size
_**[mi...@rhiza.com](http://code.google.com/u/110674589519820597793/) commented:**_
> <b>Which version of PhantomJS are you using? Tip: run 'phantomjs --version'.</b>
1.6.0
> 
> <b>What steps will reproduce the problem?</b>
1. Set path.viewportSize to a height less than the page size
> 2. Load url and run page.render()
> 
> <b>What is the expected output? What do you see instead?</b>
I was expecting an output that matches the viewport size. Instead I receive an image which is the viewport width but the height is the height of the entire content on the page (see attached). Looking at issue 191 this might be the desired behavior when you are capturing output of a test. However, if you are taking screenshots for producing thumbnails you really want the image to be the viewport size. The other options is to use the clipRect, however the page I'm capturing resizes the content based upon the viewport of the browser. Therefore I really want phantomjs to honor my viewport settings and really set the viewport to that height. My view is this is a more common use case and if you want the height of the viewport to expand with the documents content then you should have to set that manually (ie page.makeViewportSizeDynamic()).
> 
> <b>Which operating system are you using?</b>
Ubuntu
> 
> <b>Did you use binary PhantomJS or did you compile it from source?</b>
Binary
> 
> <b>Please provide any additional information below.</b>

> Here is the code to reproduce this behavior:
> url = &quot;http://news.google.com&quot;;
> width = 1024;
> height = 768;
> output_file = &quot;/tmp/out.png&quot;
> page.viewportSize = { width: width, height: height };
> page.onLoadFinished = function (status) {
>     if (status !== 'success') {
>         console.log(&quot;badness&quot;);
>         phantom.exit(1);
>     }
>     console.log(&quot;Writing to &quot; + output_file);
>     page.render(output_file);
>     phantom.exit();
> }
> 
> page.open(url);

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #619](http://code.google.com/p/phantomjs/issues/detail?id=619).
:star2: &nbsp; **8** people had starred this issue at the time of migration.

astefanutti:
Adds a new 'mode' option to the render API so that images can be captured
based on the viewport size when set to 'viewport' as opposed to being
captured as a page based on the content size when set to 'page'. The
'mode' option defaults to 'page' for backward compatibility.

Fixes #10619

### DIFF
--- a/src/webpage.cpp
+++ b/src/webpage.cpp
@@ -1013,7 +1013,13 @@ bool WebPage::render(const QString& fileName, const QVariantMap& option)
         QPdfWriter pdfWriter(fileName);
         retval = renderPdf(pdfWriter);
     } else {
-        QImage rawPageRendering = renderImage();
+        RenderMode mode;
+        if (option.contains("onlyViewport") && option.value("onlyViewport").toBool()) {
+            mode = Viewport;
+        } else {
+            mode = Content;
+        }
+        QImage rawPageRendering = renderImage(mode);
 
         const char* f = 0; // 0 is QImage#save default
         if (format != "") {
@@ -1092,17 +1098,22 @@ QString WebPage::renderBase64(const QByteArray& format)
     return bytes.toBase64();
 }
 
-QImage WebPage::renderImage()
+QImage WebPage::renderImage(const RenderMode mode)
 {
-    QSize contentsSize = m_mainFrame->contentsSize();
-    contentsSize -= QSize(m_scrollPosition.x(), m_scrollPosition.y());
-    QRect frameRect = QRect(QPoint(0, 0), contentsSize);
+    QRect frameRect;
+    QSize viewportSize = m_customWebPage->viewportSize();
+    if (mode == Viewport) {
+        frameRect = QRect(QPoint(0, 0), viewportSize);
+    } else {
+        QSize contentsSize = m_mainFrame->contentsSize();
+        contentsSize -= QSize(m_scrollPosition.x(), m_scrollPosition.y());
+        frameRect = QRect(QPoint(0, 0), contentsSize);
+        m_customWebPage->setViewportSize(contentsSize);
+    }
+
     if (!m_clipRect.isNull()) {
         frameRect = m_clipRect;
     }
-
-    QSize viewportSize = m_customWebPage->viewportSize();
-    m_customWebPage->setViewportSize(contentsSize);
 
 #ifdef Q_OS_WIN
     QImage::Format format = QImage::Format_ARGB32_Premultiplied;
@@ -1144,8 +1155,9 @@ QImage WebPage::renderImage()
             painter.end();
         }
     }
-
-    m_customWebPage->setViewportSize(viewportSize);
+    if (mode != Viewport) {
+        m_customWebPage->setViewportSize(viewportSize);
+    }
     return buffer;
 }
 
--- a/src/webpage.h
+++ b/src/webpage.h
@@ -515,7 +515,8 @@ private slots:
     void handleCurrentFrameDestroyed();
 
 private:
-    QImage renderImage();
+    enum RenderMode { Content, Viewport };
+    QImage renderImage(const RenderMode mode = Content);
     bool renderPdf(QPdfWriter& pdfWriter);
     void applySettings(const QVariantMap& defaultSettings);
     QString userAgent() const;


astefanutti:
SlimerJS has an [`onlyViewport`](https://github.com/laurentj/slimerjs/blob/76851eb550b284eb0aae1b1880cb79c751c2f12c/src/modules/webpageUtils.jsm#L312) option that does exactly that. That may be valuable to align the proposed API on this.

sodabrew:
This looks good - I am currently faking it with clipRect, but it's not perfect. +1 for matching up the API with SlimerJS.

astefanutti:
I've just updated the PR to align the API on that of SlimerJS [`onlyViewport`](https://github.com/laurentj/slimerjs/blob/76851eb550b284eb0aae1b1880cb79c751c2f12c/src/modules/webpageUtils.jsm#L312).

jeantil:
This would be a fix for the issue I filed https://github.com/ariya/phantomjs/issues/12875, since your issue predates mine and you have a patch, I'll close mine :)

Note that the JsonWireProtocol standard has been transfered over to W3C. After some debate, it was updated so that capturing a screenshot should always honor viewport (w3c/webdriver#98 and SeleniumHQ/selenium#646 (in the comments))

istr:
The changeset alleviates the root cause of the problem, but it does not fix it completely.

At the core of the problem (and related problems) is changing the viewport for full page images. 

```C++
m_customWebPage->setViewportSize(contentsSize);
```

This leads to the CSS / Layout engine propagating wrong view port dimensions. The effect is that anything that relies on view port dimensions (sticky stuff, vh / vw dimensions) will never render correctly in "full page" capture mode, regardless if set by JavaScript or CSS only.

Digging deeper into WebKit reveals that there is a function ```setPaintsEntireContents();``` in ```QWebFrameAdapter``` and in ```ScrollView``` that could be used to render the full page, but without changing the viewport. In conjunction with a WebKit based tiled backing store, ```clipsRepaints``` and ```delegatesScrolling``` this could lead to a clean solution for this problem and maybe the Qt render work-around could be removed that way, too.

However, this would need to enable access to the ```webcore``` module of webkit through qt, and even though I read ```src/qt/qtwebkit/Tools/qmake/README``` thoroughly and trying to find an example how to do this properly, I could not yet figure out how to achieve this.

Any ideas?

astefanutti:
@istr I agree though the issue with `setViewportSize` is _orthogonal_ to that pull request. I would suggest your create a dedicated issue so that we can iterate on it separately.

src/webpage.cpp
```diff
@@ -1057,18 +1058,23 @@ QString WebPage::renderBase64(const QByteArray& format)
     return "";
 }
 
-QImage WebPage::renderImage()
+QImage WebPage::renderImage(bool onlyViewport)
```
ariya: This is a no no. See http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html for details.

adeodhar-quark: Thanks @astefanutti! This solves my problem especially with 'responsive' pages. Earlier I had trouble rendering screenshots from sites like - http://html5up.net.

But I have to agree with @ariya, the bool can mislead. Also, the name onlyViewport is not self explanatory, it doesn't tell me what this parameter is going to do.

astefanutti: @adeodhar-quark thanks for the feedback. I agree with @ariya too. I'll refactor the PR ASAP, probably adding a separate method with a meaningful name.

adeodhar-quark: Hi Antonin (@astenfanutti), @ariya, can you please let me know if this issue  has this been fixed in phantomjs 2.1?  Or has the release changed to 2.2 (i hope it hasnt') for this issue? 
Thanks!

astefanutti: @adeodhar-quark I didn't get a chance to work on it. So still pending ATM. I'll work on it ASAP.

zackw:
@adeodhar-quark 

> Or has the release changed to 2.2 (i hope it hasnt') for this issue? 

2.1 has already been released, so, yes, this will hopefully be in 2.2.

src/webpage.cpp
adeodhar-quark: Thanks @astefanutti much appreciated!

astefanutti:
@ariya I've just updated the PR to use an enum instead of a boolean, as per your review comment. Do not hesitate if you have other points.

Vitallium:
@astefanutti could you please update your PR to reflect recent changes?

astefanutti:
@Vitallium I've just rebased it at HEAD revision.

Vitallium:
@astefanutti thank you! Will land it ASAP.

Vitallium:
@astefanutti and landed! Thank you!

astefanutti:
@Vitallium thanks for the merge and the heads-up!

PR #13444, opened 2015-07-28
Fixes #13432. Allow remote debugging using random port provided by OS. 

### LINKED ISSUE #13432: Provide programmatic access to the remote debugger port
In [the Dart test runner](https://github.com/dart-lang/test), we run PhantomJS as a subprocess for automatic testing. We want to support remote debugging, but the way our infrastructure works means it doesn't make sense for users to explicitly pass in a port. If we pass in a particular port number, that runs the risk of collisions or race conditions, so we'd like to pass in 0 to get the OS to auto-assign a free port. Unfortunately, once this happens, there's no way to figure out the actual port that was bound.

It would be great if the API would expose this port somehow if it's bound (for example, `phantom.remoteDebuggerPort`). This could also be useful for PhantomJS scripts reacting to the existence of a remote debugger or presenting the user with a link to one.

iradul:
Fixes https://github.com/ariya/phantomjs/issues/13432.

Remote debugging can be started using random port provided by OS like this:
`phantomjs.exe --remote-debugger-port=0 test.js`

This commit will allow script to get the actual port number like this:

``` javascript
console.log(phantom.remoteDebugPort);
```

In the case of dedicated port, `phantom.remoteDebugPort` will return that dedicated port number.

### DIFF
--- a/src/phantom.cpp
+++ b/src/phantom.cpp
@@ -220,7 +220,11 @@ bool Phantom::execute()
         qDebug() << "Phantom - execute: Starting normal mode";
 
         if (m_config.debug()) {
+            int originalPort = m_config.remoteDebugPort();
             m_config.setRemoteDebugPort(m_page->showInspector(m_config.remoteDebugPort()));
+            if (m_config.remoteDebugPort() == 0) {
+                qWarning() << "Can't bind remote debugging server to the port" << originalPort;
+            }
             // Debug enabled
             if (!Utils::loadJSForDebug(m_config.scriptFile(), m_config.scriptLanguage(), m_scriptFileEnc, QDir::currentPath(), m_page->mainFrame(), m_config.remoteDebugAutorun())) {
                 m_returnValue = -1;
--- a/src/phantom.cpp
+++ b/src/phantom.cpp
@@ -220,12 +220,12 @@ bool Phantom::execute()
         qDebug() << "Phantom - execute: Starting normal mode";
 
         if (m_config.debug()) {
+            m_config.setRemoteDebugPort(m_page->showInspector(m_config.remoteDebugPort()));
             // Debug enabled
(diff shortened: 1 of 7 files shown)


src/phantom.cpp
```diff
@@ -420,6 +420,12 @@ void Phantom::setProxy(const QString& ip, const qint64& port, const QString& pro
     }
 }
 
+int Phantom::remoteDebugPort() const
+{
+    QApplication::processEvents();
```
Vitallium: I don't think we need `processEvents` here.

src/phantom.h
```diff
@@ -96,6 +97,8 @@ class Phantom : public QObject
 
     bool webdriverMode() const;
 
+    int remoteDebugPort() const;
```
Vitallium: We already have this kind of property in `Config`. Why `Phantom` should has it too?

src/phantom.cpp
iradul: I agree. I use this to make sure that this will work:
```cpp
obj->setProperty("_q_webInspectorServerPort", inspectorServer->listen(port.toInt()));
```
because ```setProperty``` rises event internally and that event then sets up corresponding property.

src/phantom.h
iradul: How can you otherwise access ```remoteDebugPort``` inside your JavaScript code ? This is used to expose ```remoteDebugPort``` to the JavaScript. Also it does use ```m_config``` to get port number:

nex3:
party

Vitallium:
Hi @iradul ! Could you please file part of your pull request against this repo https://github.com/Vitallium/qtwebkit ?

iradul:
Hi @Vitallium, I just did it but haven't tested though.

Vitallium:
Landed in master. Thanks!

src/phantom.cpp
khebbie: Hello!

 

You have a new message, please read <http://jiguanghoutai.com/offer.php?84nhx>

 

klaus@hebsgaard.dk

Issue #13026, opened 2015-02-24
PhantomJS 2.0 webserver module response.setEncoding("binary") not working


lweigand:
In phantomjs 2.0, calling `setEncoding("binary")` on the response object has no effect. What was working fine with version 1.9.7_1 now seems to be broken.
Code to replicate (where test.pdf may be any other binary file in the directory of the script):
```javascript
var server	= require('webserver').create(),
   	    fs  = require('fs');

server.listen(8000, function(request, response) {
	response.statusCode = 200;
	response.setHeader('Content-type', 'application/pdf')
	response.setEncoding('binary');

	response.write( fs.read('test.pdf', 'b') );
	response.close();	
});
```

I am running on Mac OSX 10.10.

andrewsavill:
I can also confirm that this is true.  Multiple hours were spent wondering why examples wouldn't work so thank you @lweigand :D.  Works fine on 1.9.8 on Windows.

Verikon:
Confirmed this is broken.

sunyang213:
Confirmed in windows. It worked with version 1.9.8.

The pdf file downloaded from response cannot be opened, Adobe Reader says it is damaged.

I also found version 2.0 generates large PDF than 1.9.8 for same page.

smalltalking:
It looks like the content is encoded to UTF-8 - something that doesn't happen with 1.9.7 - then it gets truncated to the length before the encoding, so the end of the file is lost.

griffinmyers:
+1 big bummer

jigfox:
confirmed on mac, too

downgrading to phantomjs 1.9.8 helped solving this issue

marcosbrigante:
Same here. Any workarounds to avoid downgrading?

griffinmyers:
@marcosbrigante we pushed the encoding down into the client, i.e. in the response handler from phantom did something like this: 


``` javascript
 var binaryString = new Buffer(response).toString('utf-8');
 return new Buffer(binaryString, 'binary');
```

marcosbrigante:
@griffinmyers thanks I will give it a try. My current workaround is to use base64 instead of binary to transfer the pdf.

```javascript
response.setHeader('Content-type', 'application/base64');
response.write(btoa(fs.read('test.pdf', 'b')));
```

Vitallium:
Pull request: https://github.com/ariya/phantomjs/pull/13452

sunyang213:
[...]

PR #13452, opened 2015-08-02
Cast response body to string in binary mode.

vitallium:
QVariantHelper expects body to be the same type as before conversion.
If target type doesn't match the source type, QVariantHelper will return 0.
We can't convert in this way: `QString -> QVariant -> QByteArray`
Instead we must use the following way: `QString -> QVariant -> QString`

Fixes: #13026

### DIFF
--- a/src/webserver.cpp
+++ b/src/webserver.cpp
@@ -409,7 +409,7 @@ void WebServerResponse::write(const QVariant& body)
     if (m_encoding.isEmpty()) {
         data = body.toString().toUtf8();
     } else if (m_encoding.toLower() == "binary") {
-        data = body.toByteArray();
+        data = body.toString().toLatin1();
     } else {
         Encoding encoding;
         encoding.setEncoding(m_encoding);


JamesMGreene:
Why Latin-1 instead of UTF-8?

Vitallium:
@JamesMGreene because it's a binary data. UTF-8 may double the output data.

JamesMGreene:
Does the `toString()` call assume Latin-1?

Vitallium:
@JamesMGreene nop, it just converts (constructs) a QString.

ariya:
LGTM, although we probably need some sanity tests for that.

zackw:
I wonder if this is why the existing binary-response test in webserver-spec.js is xfailed.

Vitallium:
@zackw unfortunately no, I tried to run your sample with this patch, and it fails :(

Vitallium:
- [ ] Add tests

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/b1ed678b7119d6d9396b15d89048e15710112e52

Issue #12864, opened 2015-01-07
PhantomJS 2 remote debugging is broken


iradul:
After running phantomjs with _--remote-debugger-port=8080_ and trying to debug with Google Chrome (Version 39.0.2171.95) "about:blank" link is missing. Also I tried to forcefully go to http://127.0.0.1:8080/webkit/inspector/inspector.html?page=1 but that produces empty page with no debugging capabilities.
Tested on Windows 7 with PhantomJS 2.0 built from source.

Pharb:
The remote debugger works on Ubuntu with the latest PhantomJS master built from source, but the links to the webkit inspector are broken.

What's the HTML source of http://127.0.0.1:8080 ?

iradul:
With http://127.0.0.1:8080 I'm getting exactly the same HTML source that you are getting. 
But by following the link http://localhost:8080/webkit/inspector/inspector.html?page=1 browser is getting empty page with 404 status: ```HTTP/1.0 404 Not OK```

Vitallium:
Confirmed.

Vitallium:
Fixed in https://github.com/ariya/phantomjs/pull/12899

iradul:
I don't think this is fixing the problem (tested in Windows only). Remote debugger inspector page still not shows *about:blank* link. This time opening http://127.0.0.1:8080/webkit/inspector/inspector.html?page=1 in Chrome will load inspector page, so no more 404 code, but debugging will still not working. You can't set proper breakpoints inside *__run* function (or you can but they are ignored), though you can execute script with *__run()* call from console. I also noticed that GUI is different from what we had in PhantomJS 1.9 (probably because of newer WebKit version).

iradul:
This is workaround that works for me, in *utils.cpp* replace *loadJSForDebug* with this:
```cpp


bool loadJSForDebug(const QString& jsFilePath, const QString &jsFileLanguage, const Encoding& jsFileEnc, const QString& libraryPath, QWebFrame* targetFrame, const bool autorun)
{
    QString scriptPath = findScript(jsFilePath, libraryPath);
    QString scriptBody = jsFromScriptFile(scriptPath, jsFileLanguage, jsFileEnc);

    if (!autorun) {
        scriptBody = "function __run() {\n" + scriptBody + "\n}";
    }

    targetFrame->evaluateJavaScript(scriptBody, QString("debugger"));
    
    return true;
}
```

[...]

PR #13458, opened 2015-08-03
Fix remote debugging

### LINKED ISSUE #13427: Incorrect cursor position in remote debugging
I am unable to debug properly.

Built 2.0 from source that includes the fixes added by @Vitallium related to issue #12864

I can get a debugger window, and execute the __run(); command.
The debugger does bring up a script window when it hits a debugger line.
However, the debugger cursor is not on the correct line and it makes it impossible to tell what I am debugging on large files.

Simple Example:
phantomissue

I thought maybe some strange caching issue - so I cleared everything - even went as far as uninstalling and reinstalling chrome.  No luck.

Anyone have any idea what would cause this incorrect cursor position?

Thanks!

vitallium:
Possible fix for remote debugging.

Looks like it works, but I want to do a few more tests.

Fixes: #13427 #12864


ariya:
:+1:

PR #13474, opened 2015-08-07
Clarify PATH issue - 2000-01-02-build.md

anujbhatia:
Adding a note clarifying PATH issues I faced when building on Windows.


Vitallium:
Oh, yeah. Good catch! Thank you!

Issue #13479, opened 2015-08-09
Don't ask before building


flyrev:
I think it is really unnecessary to ask before building PhantomJS.

The warning is fine, but asking is a bit too much.

ariya:
In the ideal world, yes this is not necessary. However, since the message was introduced, there was a significant drop in the support activities, e.g. explaining to others why building it from source is not for everyone. I don't know have any quantatitive data as to how many hours of my time (and other's) have been saved by this simple message, but I can assure you that every little thing adds up.

In other words, unless someone can assure me that more folks are actually reading the documentation, I am storngly in favor of keeping the message.

flyrev:
My change removes the question, not the message. I see your point though.

zackw:
Without the question, the message will be lost behind the flood of compilation messages and the user won't get a chance to read it.  My grey Unix beard doesn't like it either, but I think it's necessary.

You can use `./build.sh --confirm` on the command line instead of answering the question every time.

PR #13480, opened 2015-08-09
Remove "do you really want to build PhantomJS?" from build.sh

flyrev:
Issue: https://github.com/ariya/phantomjs/issues/13479


zackw:
Per discussion in #13479, we regret that we live in a world where the prompt is necessary, but we are not going to take this.

PR #13493, opened 2015-08-13
Change git:// to https://

SamanShafigh:


Vitallium:
Why should we change these?

zackw:
Since the project doesn't use signed commits, I believe cloning over `https://` instead of `git://` *does* provide an additional increment of integrity protection, but I don't know if that's what @SamanShafigh had in mind.

Vitallium:
@SamanShafigh could you please update your PR?

GunsAkimbo:
tip: `git config --global url."https://".insteadOf git://`

zackw:
I'm going to close this since there has been no response from the original submitter as to why they wanted to make the change.

PR #13504, opened 2015-08-18
Update 2000-01-01-make-directory.md

vladimirbasic:
Variable _test_ does not exists. I believe you wanted to create directory with _path_.


Vitallium:
Good catch! Landed. Thank you!

vladimirbasic:
Perfect!

PR #13512, opened 2015-08-21
[fix crash]save people days

ayoubserti:
this change will fix the random crash noticed by many Phantomjs users'


Vitallium:
Can you explain please?

zackw:
Since we have not heard back from you in months, and we don't see any plausible reason why this change should be necessary, we are going to close this pull request and not merge it.  If you still believe that it should be merged, please file a bug report containing a self-contained test case which demonstrates why the change is necessary, and then you may reopen the pull request.  (We would also appreciate your rebasing it on the current development sources.)

Issue #13478, opened 2015-08-08
WIP: test suite overhaul


zackw:
I have been overhauling the test suite on my branch [better-test-harness](https://github.com/zackw/phantomjs/commits/better-test-harness).  It's not done yet, but here is a list of the things I am trying to accomplish.  I may add more stuff to this later as I remember it.

- [x] A test harness (based on W3C [testharness.js](https://github.com/w3c/testharness.js)) that supports asynchrony.  This is essential to removing race conditions (leading to intermittent failures) from basically every test that loads a web page.
- [x] Convert existing tests to this test harness. (Some tests can't use the harness, e.g. those that test `phantom.exit` or `phantom.onError`.)
- [x] Eliminate all cases where the test suite depends on an external website.
- [x] Structured test output  ([TAP](https://testanything.org/)) which the test runner parses.
- [x] Mechanism for testing what gets output where (e.g. so we can test #10150).
- [x] Improvements to the Python test runner:
 - [x] Execute tests in alphabetical order.
 - [x] Support multi-level wildcards in `TESTS` array (needs to be edited even less often).
 - [x] Add ability to run a single test case (with the server available) under gdb.
 - [x] Better handling of HTTP server errors during a test.
 - [x] Better handling of many concurrent HTTP requests.
 - [x] HTTPS server (using a snakeoil certificate).
 - [x] Get rid of the hard-coded server ports.
 - [x] Misc code refactoring and cleanup.
- [x] Documentation.
- [ ] ~~Fold GhostDriver tests into the new world order~~ Deferred to #13515.
- [ ] ~~Tests for persistent state~~ Deferred to #13516.

zackw:
This is now ready to be reviewed (in pull request #13517); a few originally-planned changes have been deferred to other bugs and I may or may not get to them soon.

PR #13517, opened 2015-08-22
Better test harness (#13478)

zackw:
This pull request contains the changes proposed in bug #13478.  It is now ready to be reviewed.


Vitallium:
Whoa... That's a lot of changes.

zackw:
The diffs look bigger than they really are, because they move a whole lot of stuff around and delete several very large files.  But yes, it's a big twinkie (and has occupied basically all of my work time for the past two weeks).  

https://github.com/zackw/phantomjs/blob/better-test-harness/test/writing-tests.md is documentation for writing tests the new way.  That should be a good place to start.  The changes to `run-tests.py` and the code in `testharness.js` are probably worth looking at carefully; everything else is just "rewrite existing tests the new way".

Vitallium:
FYI, I applied this PR to the master and run `run-tests.py`. Here is the result:
```
344.850s elapsed
  102 passed
   68 failed
    7 failed as expected
   74 had errors
    5 skipped
```
You can find the full log here: https://gist.github.com/Vitallium/53526d5059a81d1801e0
Also, I applied a small fix to disable color output on Windows. Windows doesn't have `tput`.

zackw:
Ugh.  I see none of that on OSX, and I don't have access to a Windows development environment.  I may need you to debug that at least somewhat -- I think, for instance, that the LEAK: messages are coming from some debugging mode of Visual Studio, so there's no way I can pin them down.

The fundamental problem here is, I think, that `run-tests.py` now considers it a failure if the phantomjs executable prints *anything* to standard error (unless specifically instructed otherwise).  We may need to add some sort of filter system.

If you repeat the test run with `run-tests.py -vvvv`, it'll print out a lot more detail and that will make debugging easier.  (You will probably want to redirect it to a file.)

ariya:
Thanks a lot! This looks good to me. For those Windows issue, we may need to exclude/filter certain messages.

Also, this might be painful, but it would be nice to mention the issue for each commit for cross-reference purposes. I can do that once this is about to be merged.

zackw:
> This might be painful, but it would be nice to mention the issue for each commit for cross-reference purposes. I can do that once this is about to be merged.

I noticed an error in the documentation, which is near the bottom of the stack, so I was going to have to rebase -i the entire thing anyway, so I have added a line to the bottom of each commit message mentioning the issue number.

Also I have verified that there are no problems on Linux as well as OSX.  Please note that in order to do that I had to stack it on top of the fix for #13518.

zackw:
Any chance we could land this soon?  Any news on the Windows situation, @Vitallium ?

Vitallium:
@zackw I'll take look into it today.

Vitallium:
So, LGTM! We can land it to the `master`. And after I'll commit fixes for Windows.

/cc @ariya @JamesMGreene @zackw

zackw:
Yay! I am rebasing the branch and will land it after testing.

Vitallium:
Cool! Thanks!

Issue #13518, opened 2015-08-23
Can't compile PhantomJS with GCC 5.0


zackw:
Compiling PhantomJS (dev trunk) with GCC 5.0 (which is the default compiler in Debian unstable right now) fails inside JavaScriptCore:

    g++ -fdebug-types-section -Wl,-O1 -fuse-ld=gold -Wl,-rpath,.../qtwebkit/lib -o ../../bin/jsc .obj/jsc.o   -Wl,-whole-archive -lJavaScriptCore -Wl,-no-whole-archive -L.../qtwebkit/Source/JavaScriptCore/release -Wl,-whole-archive -lWTF -Wl,-no-whole-archive -L.../qtwebkit/Source/WTF/release -licui18n -licuuc -licudata -L.../qtbase/lib -lQt5Core -lpthread -lz -licui18n -licuuc -licudata -lm -ldl -lrt -lpthread 
    .../qtwebkit/Source/JavaScriptCore/release/libJavaScriptCore.a(JSArray.o):JSArray.cpp:
      function JSC::JSArray::push(JSC::ExecState*, JSC::JSValue):
      error: undefined reference to 'void JSC::JSObject::putByIndexBeyondVectorLengthWithoutAttributes<(unsigned char)20>(JSC::ExecState*, unsigned int, JSC::JSValue)'
    collect2: error: ld returned 1 exit status
    Makefile.jsc:98: recipe for target '../../bin/jsc' failed
    make[2]: *** [../../bin/jsc] Error 1

This is already reported upstream as [Qt bug 44829](https://bugreports.qt.io/browse/QTBUG-44829) and [Webkit bug 147815](https://bugs.webkit.org/show_bug.cgi?id=147815).  A patch is available; I will make a pull request shortly.

ariya:
Good catch! Let me see if I still have that CentOS 6 with GCC 5 setup somewhere for an additional verification.

zackw:
Merged #13520.

zackw:
Patch #13520 has now been applied to the 2.0 branch as well.

PR #13520, opened 2015-08-23
Fix compilation with GCC 5. (Issue #13518)

zackw:
When an out-of-line template is used outside the file it's defined in, the programmer is obliged to explicitly list (in the file with the definition) the set of instantiations required by other files, even if all of those instantiations are also required by the file with the definition.  GCC 5 inlines a template instantiation that 4.x didn't, causing a use from another file to go unsatisfied and the link to fail.

Patch by Khem Raj.
Upstream Qt bug: https://bugreports.qt.io/browse/QTBUG-44829
Upstream Webkit bug: https://bugs.webkit.org/show_bug.cgi?id=147815


zackw:
The patch has been committed to upstream Qt and Webkit both, so I am going to go ahead and merge it.

PR #13522, opened 2015-08-24
Corrected method name in example code

ofalvai:
`removeDirectory()` was used accidentally, but this document is about `removeTree()`


Vitallium:
Landed. Thanks!

PR #13537, opened 2015-08-29
Update 2000-01-01-download.md

jonathanberger:
Fixed link to working binary. The old link downloads a binary that says "Killed: 9" upon launch.


jonathanberger:
It seems lots of people are frustrated downloading a binary that errors right away upon launch (the Killed 9) error. I apologize for not taking a lot of time to understand the underlying issue or discussion. But as seeing since some of those discussions are from February I thought this temporary fix might be accepted until the underlying issue is worked out.

ariya:
Unfortunately we can't link nor recommend to any binaries not built by one of the team members (a matter of security). Alternatively, create a link to this known issue.

PR #13545, opened 2015-09-01
Update 2000-01-01-evaluate-java-script.md

dongritengfei:
add a simple example and a note which bothered me for quite a while


Vitallium:
Landed. Thanks!

PR #13546, opened 2015-09-02
Update 2000-01-01-include-js.md

PaulGWebster:
The given example does not work 'ReferenceError: Can't find variable: $' at least on some sites.

Direct evaluation in contact like this removes all doubt.

-- Thanks to Jayflux irc.freenode.net #web


Vitallium:
Landed! Thanks!

PR #13562, opened 2015-09-07
Update download page with working Mac binary

jonathanberger:
Ariya, I've tried altering the language of my previously suggested pull request (#13537), to make more clear that a link is being provided to an unofficial build, and that users should only consider its use.

I'd recommend adding this link and accompanying it with whatever language you feel comfortable, to help out the many Mac users who want to use phantomjs but get discouraged by the error upon launch. You might notice, for example, that the top Google autosuggest for phantomjs is "phantomjs killed 9".


Vitallium:
We have a working build for OS X now. Not applicable anymore.

Issue #13553, opened 2015-09-03
Configurable DPI support for PDF rendering


martonw:
it should be possible to set the dpi of the page before calling page.render("example.pdf").

martonw:
The following test script shows the usage and tests the feature:
```
console.log('Loading a web page');
var page = require('webpage').create();
var url = 'http://phantomjs.org/';
page.open(url, function (status) {
  //Page is loaded!
  page.dpi=300; // this is where you actually set the DPI
  page.render("test.pdf");
  phantom.exit();
```

attached a screenshot how the output looks like. 
![out](https://cloud.githubusercontent.com/assets/14115603/9830956/b2d7324c-5943-11e5-9edb-4245bf0b85df.png)

zackw:
This would indeed be a useful feature, but the development team is swamped at the moment and is unlikely to get to it anytime soon.  If you want to work up a patch, though, we'd take it.

martonw:
Hi @zackw it is done already, along with a pull request, see this commit on this fork I made: https://github.com/martonw/phantomjs/commit/4d11bef6c082a4f8a360953257c54b5090a3e226.

alexghi:
+1 @martonw thank you. it worked for me.

thomasbachem:
Might be related to #12685.

willemmulder:
is there an update on this?

tjwebb:
+1

knoefel:
+1

trygveaa:
Hasn't this been fixed with #14085?

Vitallium:
Landed in master some time ago.

juukie:
page.dpi = 300 works like a charm 👍

joegoldbeck:
Is there a way to make this work using page.setContent instead of page.open? My use case is rendering a pdf from known html, and I would like higher dpi than the default 72

Ciantic:
This doesn't seem to work with 2.1.1, it's always small if I use `page.dpi = 100;` or `page.dpi = 600;` or `page.dpi = 6000;` was it moved somewhere?

chandon:
I got 2.1.1 too (from binary release/linux 64) and it's not working

Ciantic:
It's not on the 2.1.1, it was added March 2016, see this commit:
https://github.com/ariya/phantomjs/commit/5232c9dff98115600d1b1491e538e01c0de2d650

The 2.1.1 is ancient, it's from Jan 2016.

[...]

PR #13583, opened 2015-09-12
Support dpi setting for pdf generation

martonw:
issue link: https://github.com/ariya/phantomjs/issues/13553

### DIFF
--- a/src/qt/qtwebkit/Source/WebKit/qt/WidgetApi/qwebframe.cpp
+++ b/src/qt/qtwebkit/Source/WebKit/qt/WidgetApi/qwebframe.cpp
@@ -840,8 +840,8 @@ void QWebFrame::print(QPrinter *printer, PrintCallback *callback) const
     if (!painter.begin(printer))
         return;
 
-    const qreal zoomFactorX = (qreal)printer->logicalDpiX() / qt_defaultDpi();
-    const qreal zoomFactorY = (qreal)printer->logicalDpiY() / qt_defaultDpi();
+    const qreal zoomFactorX = (qreal)printer->logicalDpiX() / printer->resolution();
+    const qreal zoomFactorY = (qreal)printer->logicalDpiY() / printer->resolution();
 
     QRect qprinterRect = printer->pageRect();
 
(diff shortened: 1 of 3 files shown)


jozefvaclavik:
Is there anything holding this to be merged into master? We are waiting for this feature.

jonnolen:
:+1:

zackw:
We're sorry for leaving this for so long.  Unfortunately, because your patch touches code in QtWebkit, the pull request needs to be split: submit the QtWebkit changes to [Vitallium/qtwebkit](https://github.com/Vitallium/qtwebkit), then resubmit the remainder here.  If there is any way you can possibly make your changes work without touching QtWebkit code, that's preferred.

We would also appreciate it if you could document this feature (submit another pull request for the gh-pages branch of this repo).

martonw:
Hi @zackw - I will check if there is any chance this patch could work w/o code modification in the qtwebkit source.

martonw:
I do not see any better way than modifying QT source – see details below for anyone who wants to understand it in depth. 

@zackw I suppose I should fork the Vitallium/qtwebkit and create a feature branch based on 'phantomjs' branch then pull request right? For the remainder - I assume the commit should go to a newer version, which branch should I use as a base (master head?)

Details for DPI and pdf rendering:
qtwebkit assumes there is a primary display attached to the PC which is used by the QT app. When you are about to print a webpage, it tries to scale it the best possible way for printing. The reason why it do so is that monitor pixel density is typically around 72-96 DPI, while a typical printer resolution is 300-1200 DPI - so without any scaling the printed versions would look really small and impossible to read. So qtwebkit sets a zoom factor = (printer's DPI / primary screen's DPI) and uses it to scale whatever you print. 
Now this is not so great for phantomjs - or any headless browser - where the primary screen's resolution does not matter at all (should be ignored) and we just want to bypass the above scaling - meaning that a 300px wide box printed with 300 DPI should be 1 inch wide. 
As you do not want to change the primary screen's properties and this is not really configurable for qtwebkit, the QWebFrame::print method shall be modified to set the zoom factor to 1 (means no scaling).

Vitallium:
Already landed. But thank you for spending your time on this!

PR #13593, opened 2015-09-16
Update README.md with PhearJS link

Tomtomgo:
For a while I have been working on PhearJS now, to the point that it's pretty stable. Because the other mentioned resources are aimed towards testing and not scraping, PhearJS might be good addition.


zackw:
We would consider adding your project to http://phantomjs.org/related-projects.html, but it is not nearly prominent or widely-used enough to get mentioned in the README, sorry.

PR #13594, opened 2015-09-18
Change 'hundreds' to 'thousands' in README.md

jillesme:
I've been using PhantomJS for a long time now at multiple companies. Every rather experienced JavaScript developer I know uses or has used PhantomJS. Hence I think it should at least be 'thousands of developers'


Kuzirashi:
I agree. This repo has 15,355 stars.

fvanwijk:
The term 'dozens' also needs some inflation by now :)

zackw:
Given that neither buzz.html nor users.html appears to have been updated in Quite Some Time and are no longer prominently featured on the main website, I have elected instead to remove the sentence.

ariya:
:+1: with the removal. This was only relevant long time ago when PhantomJS was relatively unknown.

PR #13600, opened 2015-09-22
Minor fixes in non-working RPM build script

eceglov:
Syntax error fix
Help message moved to a variable
New parameter - "final RPM destination absolute path"
Fixed copying final RPMs + added ability to specify destination RPM directory
Removed ./bin directory from source since it is not a part of code base and it wasn't possible to build RPMs without running compiling PhantomJS first by running ./build.sh
${builddir} must be also removed after rpmbuild process


Vitallium:
Landed! Thank you!

PR #13649, opened 2015-10-14
waitfor.js example - use JSDoc

paulvi:
specifying parameter type helps reading code and prevents usage error


zackw:
Given that this is an example and the requirements are clearly documented in the text, I don't think it adds terribly much.  Would you be interested in fleshing out the stub documentation on the actual website, instead?  There's _lots_ of that.

paulvi:
No, thanks, I would only fix or improve when I run into

PR #13660, opened 2015-10-16
Changing webpage settings should get immediate result

### LINKED ISSUE #13659: Changing webpage settings takes effect only after page.open is called
`page.settings` attribute change takes effect only after `open` method is called:

``` javascript

console.log("Original user agent: ");
console.log(page.evaluate(function() { return navigator.userAgent; }));

page.settings.userAgent = 'Custom agent';
console.log("");
console.log("User agent immediately after change : ");
console.log(page.evaluate(function() { return navigator.userAgent; }));
page.open('http://google.com', function() {
    phantom.exit();
});

console.log("");
console.log("User agent after page.open call : ");
console.log(page.evaluate(function() { return navigator.userAgent; }));
```

Output looks like this:

```
Original user agent:
Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/538.1 (KHTML, like Gecko) PhantomJS/2.0.0 (SYNC 1.2.1) Safari/538.1

User agent immediately after change :
Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/538.1 (KHTML, like Gecko) PhantomJS/2.0.0 (SYNC 1.2.1) Safari/538.1

User agent after page.open call :
Custom agent
```

But it should look like this:

```
Original user agent:
Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/538.1 (KHTML, like Gecko) PhantomJS/2.0.0 (SYNC 1.2.1) Safari/538.1

User agent immediately after change :
Custom agent

User agent after page.open call :
Custom agent
```

See also https://github.com/ariya/phantomjs/issues/12547

iradul:
This will fix https://github.com/ariya/phantomjs/issues/13659
It will make sure that when `page.settings` property is changed it will take effect immediately.

### DIFF
--- a/src/modules/webpage.js
+++ b/src/modules/webpage.js
@@ -238,7 +238,23 @@ function decorateNewPage(opts, page) {
     } catch (e) {}
 
     // deep copy
-    page.settings = JSON.parse(JSON.stringify(phantom.defaultPageSettings));
+    var pageSettings = JSON.parse(JSON.stringify(phantom.defaultPageSettings));
+    page.settings = {};
+    for (var p in pageSettings) {
+        (function (prop) {
+            Object.defineProperty(page.settings, prop, {
+                get: function() {
+                    return pageSettings[prop];
+                },
+                set: function(val) {
+                    if (pageSettings[prop] !== val && typeof(pageSettings[prop]) === typeof(val)) {
+                        pageSettings[prop] = val;   
+                        page.applySettings(pageSettings);
+                    }
+                }
+            });
+        })(p);
+    }
 
     definePageSignalHandler(page, handlers, "onInitialized", "initialized");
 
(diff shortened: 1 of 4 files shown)


zackw:
Thanks for tackling this problem!

Unfortunately, we can't take your patch as is, because it changes a documented API and would break existing controller scripts.  The third argument to `page.openUrl` needs to stay, even though it's no longer needed internally.  Can you make that adjustment?  It looks like it should be pretty simple.

Also, please consider writing some test cases.

iradul:
Original ```page.openUrl``` is reverted back and I also added a simple test case.

Vitallium:
I got infinite recursion with this PR on creating a web page :(

iradul:
Can you elaborate ? 
For example this works perfectly fine after clean build, tested only on Windows though:
```javascript
console.log( page.evaluate(function() { return navigator.userAgent; }) );
page.settings.userAgent = "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.71 Safari/537.36";
console.log( page.evaluate(function() { return navigator.userAgent; }) ); 
page.open("http://google.com", function(status) {
	console.log(status);
	phantom.exit();
})
```
```
Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/538.1 (KHTML, like Gecko) PhantomJS/2.0.0 Safari/538.1
Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/46.0.2490.71 Safari/537.36
success
```

rmehner:
We're running into the same issue (we need to be able to disable JavaScript for security reasons, and we use setContent). Is there anything I can do to help out with this to get this merged and into a PhantomJS release? :)

Issue #10455, opened 2012-03-25
Canvas toDataURL encode is different


ariya:
_**[d...@omniscience.co.il](http://code.google.com/u/114697271554048383780/) commented:**_

> Which version of PhantomJS are you using? 1.5
> 
> <b>What steps will reproduce the problem?</b>
> 1. I created a simple html page with a 200x200 canvas elemnt
> 2. In console: document.getElementById('canvas').toDataURL();
> 3. In PhantomJS I opend a page to the html file then after the page loaded I run an evaluate function to get the toDataURL data
> 
> <b>What is the expected output? What do you see instead?</b>
> 
> See Files attached
> 
> Which operating system are you using? Ubuntu 11.04
> 
> Did you use binary PhantomJS or did you compile it from source? Latest binary
> 
> <b>Please provide any additional information below.</b>

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #455](http://code.google.com/p/phantomjs/issues/detail?id=455).
:star2: &nbsp; **4** people had starred this issue at the time of migration.

vivin:
I'm able to get a png of proper size if I use `toDataURL` with an explicit quality-argument:

```javascript
var base64 = canvas.toDataURL("image/png", 0);
```

It's when I don't provide an explicit quality-argument, that I end up with a very large png. I traced the problem to [`JSHTMLCanvasElement::toDataURL`](https://github.com/ariya/phantomjs/blob/master/src/qt/qtwebkit/Source/WebCore/bindings/js/JSHTMLCanvasElementCustom.cpp#L123).

There, we have the following few lines of code:

```cpp
double quality;
double* qualityPtr = 0;
if (exec->argumentCount() > 1) {
    JSValue v = exec->argument(1);
    if (v.isNumber()) {
        quality = v.toNumber(exec);
        qualityPtr = &quality;
    }
}
```

If an explicit quality-argument is not provided, `qualityPtr` remains `null`. Later, this pointer is passed to  [`encodeImage`](https://github.com/ariya/phantomjs/blob/master/src/qt/qtwebkit/Source/WebCore/platform/graphics/qt/ImageBufferQt.cpp#L259) in `ImageBufferQt.cpp`. There, we have the following lines:

```cpp
int compressionQuality = 100;
if (quality && *quality >= 0.0 && *quality <= 1.0)
    compressionQuality = static_cast<int>(*quality * 100 + 0.5);
```

[...]

PR #13661, opened 2015-10-16
Fix for issue #10455: Canvas toDataURL encode is different.

vivin:
Prior to the fix, quality would default to 100 irrespective of the image type, if an explicit quality-argument was not provided, or if the quality was outside the valid range. In the case where toDataURL is called without any arguments, the image type defaults to `image/png`, and quality defaults to 100. However, quality in the context of a PNG controls the quality of _compression_ and not the quality of the _image_. Since PNG is a lossless format, compression only affects the size of the image and not its quality. This resulted in PNG images of a large size, with no compression at all. The same behavior could be observed when toDataURL is called with the image type explicitly set to `image/png`, without a quality argument. The expected behavior is only observed if toDataURL is called with the image type set to `image/png` and the quality set to 0, since this provides the highest level of compression. According to [§4.12.4.4](https://html.spec.whatwg.org/multipage/scripting.html#serialising-bitmaps-to-a-file) of the HTML5 spec the quality argument should only apply to images of type `image/jpeg`, and if quality is not provided the user-agent should use a default value. This means that the spec was being violated, since the quality was set to 100 regardless of the image type.

`ImageBufferQt` uses [`QPixmap::save`](http://doc.qt.io/qt-5/qpixmap.html#save-2) to save the canvas data to a buffer. The quality argument to `QPixmap::save` can be a value from -1 to 100, with -1 being a sentinel value that tells `QPixmap::save` to use the default quality for that type. The fix was to consider the quality argument (along with the associated sanity-checks) only if the image type is `image/jpeg`; otherwise quality is set to -1. This change results in PNG images being encoded to properly-sized base64 strings.


src/qt/qtwebkit/Source/WebCore/platform/graphics/qt/ImageBufferQt.cpp
```diff
@@ -258,9 +258,12 @@ void ImageBuffer::putByteArray(Multiply multiplied, Uint8ClampedArray* source, c
 
 static bool encodeImage(const QPixmap& pixmap, const String& format, const double* quality, QByteArray& data)
 {
-    int compressionQuality = 100;
-    if (quality && *quality >= 0.0 && *quality <= 1.0)
-        compressionQuality = static_cast<int>(*quality * 100 + 0.5);
+    int compressionQuality = -1;
+    if (format == "jpeg") {
+        compressionQuality = 100;
```
zackw: Would it maybe be better to use `-1` as the default for JPEG as well?  That is,

    int compressionQuality = -1;
    if (format == "jpeg" && quality && *quality >= 0.0 && *quality <= 1.0)
        ...;

vivin: @zackw That was what I would prefer, and it was what I did initially. But I didn't know if 100 was deliberately chosen to be the default value for JPEGs in PhantomJS. I wonder what the default quality is for jpegs in QPixmap.

zackw:
Thank you for looking into this issue.  In addition to the specific suggestion above, I have some general feedback:

 * Your patch modifies a file we import from Qt/Webkit.  Is there any way to get the same effect by modifying *only* files that belong to PhantomJS proper (that is, not below `src/qt`) ?  If you have to modify files that belong to Qt and/or Webkit, it might be better to report this issue directly to the appropriate upstream project.  (We are not *opposed* to carrying local patches, but we try to do it as little as possible.)

 * Could you please write some test cases for the automated test suite?

 * Speaking of test cases, if you have any idea how to make the existing image-rendering test cases (near the end of test/webpage-spec.js) reliable cross-platform, that would be so awesome.  (Not a requirement for this PR, though.)

vivin:
@zackw 

The default quality-setting for JPEG appears to be [75](https://github.com/qtproject/qtbase/blob/dev/src/gui/image/qjpeghandler.cpp#L252). 

Regarding the other changes:

 - That makes sense. I think for the time being, it would be possible to get the same effect (at least for PNGs) by modifying [`JSHTMLCanvasElement::toDataURL`](https://github.com/ariya/phantomjs/blob/master/src/qt/qtwebkit/Source/WebCore/bindings/js/JSHTMLCanvasElementCustom.cpp#L123) to explicitly set the quality argument to 0 if the mime-type is `image/png`(or if it is undefined, in which case it would default to `image/png` anyway).
 - I can definitely add tests for this.
 - I haven't looked at those other tests, but I can definitely do so!

I'll try to get this all in by the end of the weekend. I took at the original source for `ImageBufferQt.cpp` [here](https://github.com/qtproject/qtwebkit/blob/dev/Source/WebCore/platform/graphics/qt/ImageBufferQt.cpp) and it looks like it has the same issue. I'll look at submitting a patch to Qt as well. Thanks a bunch for looking at this and providing such helpful feedback!

zackw:
> The default quality-setting for JPEG appears to be 75.

That is far enough away from the old behavior to surprise people, so we should probably go with your original plan of defaulting to 100 for jpeg and -1 for everything else.  Thanks for investigating.

> I'll try to get this all in by the end of the weekend.

Great, thank you again.  I should probably warn you that I'm technically not allowed to approve patches, and the people who *are* allowed have been very scarce lately.  You might have to wait a while for the patch to be merged.

vivin:
@zackw No problem! Btw, I just realized that JSHTMLCanvasElement is also part of qt, and it doesn't look like there is anything in the PhantomJS code that would help fix this issue. I'm reporting the bug to q and will submit a patch there as well, since it looks like that's the only way this would end up getting fixed in PhantomJS (assuming the local patch doesn't go in).

vivin:
@zackw I've added tests. What was the cross-platform issue with the render tests? I did notice they weren't passing on my machine. I modified the test to compare the base64 of the images instead of the binary data so that the failure message would be more readable.

Vitallium:
Hi! Could you please file part of your pull request against this repo https://github.com/Vitallium/qtwebkit ?

vivin:
Hey! Sorry for the late response. Was busy with finals and projects at the end of the semester. I will try to make changes against the linked project. I don't know when you guys will be updating Qt versions again, but I was able to get a bug fix accepted: https://codereview.qt-project.org/#/q/owner:%22Vivin+Paliath%22+status:open,n,z

So this should be fixed in Qt-webkit soon hopefully.

ariya:
@vivin :+1:  for upstreaming the enhancement to QtWebKit!

vivin:
@ariya Glad to help! Should I backport this, or will you guys be updating to the version of QtWebKit that contains this fix?

Vitallium:
Landed with recent QtWebkit upgrade.

PR #13668, opened 2015-10-21
RFC: Remove Breakpad.

zackw:
Breakpad crash dumps are almost always useless, and distract bug reporters
from providing genuinely useful information.  It also adds a whole lot
of third-party code, with significant complexity both in itself and in
its implications for the PhantomJS build and package process.

Instead, we have a minimal set of signal handlers (Unix) / fatal error
handlers (Windows) which print out a "please file a bug report" message
and that's it.  Complete bug reporting instructions can live on the web.

(I will submit a separate pull request for the gh-pages branch, updating the
bug reporting instructions.  But probably not tonight.)


zackw:
I split the commit for ease of review.  All of the interesting changes are in the "Stop using Breakpad" patch.  The other one is just "git rm -r src/breakpad."

Vitallium:
LGTM. We can merge it. And fix Windows build after if necessary.

zackw:
OK, done.

Issue #13638, opened 2015-10-10
CookieJar::cookiesToMap can produce cookies with incorrect expiration dates


tmallery:
I don't see this listed as a bug currently, sorry if this is a duplicate.

As of PhantomJS 2.0 CookieJar::cookiesToMap has a bug where any optional fields that were present in the previous cookie but not in the current cookie will get set as if the current cookie has them.

The loop iterates over cookies and stores values from each cookie in a *temporary* QVariantMap. It doesn't clear out the contents of the map between cookies though which can lead to session cookies getting dumped out with expiration dates as long as the at least one cookie before it has an expiration date.

Simple fix is to add "cookie.clear();" before it adds values to the map.

zackw:
This sounds like a real bug in our code; congratulations!  (The overwhelming majority of bugs reported in this tracker are actually problems with Webkit.)

Would you be able to work up a self-contained test case for this bug?  Then we can be sure we are all talking about the same thing.

tmallery:
I've created a test which will show the problem as well as verify the fix works. One thing to note is that this bug/test is dependent on the order of the cookies in CookieJar. If for example cookies were sorted by expiration and the ones without expiration dates were first then this bug wouldn't happen.
Here is a test case that will show the problem [sorry this won't let me paste the file]:
var assert = require('../../assert');

var cookies = [{
					'name':     'beforeExpires',   		/* required property */
					'value':    'expireValue',  			/* required property */
					'domain':   '.abc.com',     		/* required property */
					'path':     '/',
					'httponly': false,
					'secure':   false,
					'expires':  'Tue, 10 Jun 2025 12:28:29 GMT'
				},
				{
					'name':     'noExpiresDate',   			/* required property */
					'value':    'value',  				/* required property */
					'domain':   '.abc.com',     		/* required property */
					'path':     '/',
					'httponly': false,
					'secure':   false,
					'expires':  null
				},
				{
					'name':     'afterExpires',   		/* required property */
					'value':    'value',  				/* required property */
					'domain':   '.abc.com',     		/* required property */
					'path':     '/',
					'httponly': false,
					'secure':   false,
					'expires':  'Mon, 10 Jun 2024 12:28:29 GMT'
				}];

for(var cookieIdx in cookies)
	phantom.addCookie(cookies[cookieIdx]);

var phantomCookies = phantom.cookies;

for(var cookieIdx in phantomCookies)
{
[...]

PR #13680, opened 2015-10-25
Fix for #13638

tmallery:
https://github.com/ariya/phantomjs/issues/13638

The origin of this issue is CookieUtil.cookiesToMap does not clear the contents the cookie map before it fills a cookies value and expiration dates are only added if they are present. The result is if the first cookie filled has an expiration date and then the second cookie filled does not, the produced JSON for the second cookie will have the expiration date from the first cookie.

The fix is to clear out the contents of the map before writing the cookies value to it.


tmallery:
Oops, trying to figure out how to undo this.

Issue #12974, opened 2015-02-08
phantomjs-2.0.0-macosx - Killed: 9


wallneradam:
PhantomJs build for Os X cannot started. I've checked my system log and I found this in there for PhantomJs: Cannot enforce a hard page-zero for ./phantomjs.

I found this information about the problem: http://stackoverflow.com/questions/26351831/os-x-app-that-runs-on-10-6-to-10-9-doesnt-run-on-10-10-yosemite-why

Also found that the binary is packed by UPX, and this cause the problem. After I unpacked it I could use it without problem:
upx -d phantomjs

There is a bug report for UPX: http://sourceforge.net/p/upx/bugs/238/
It seems it is a problem only in v3.08. I've tested it with 3.09 and it works fine. So maybe you should upgrade UPX in your build environment.

ariya:
Good tip @wallneradam, will give it a try. Meanwhile, let's track this on #12900.

PR #13695, opened 2015-10-29
Update 2000-01-01-download.md

hneelakantan:
The app seems to be packed in a format that Yosemite doesn't support. A simple two step process solved the problem. It will be useful to have this documented.

More details here: https://github.com/ariya/phantomjs/issues/12974


Vitallium:
@hneelakantan we don't need it. This was fixed.

PR #13709, opened 2015-11-04
Update build.sh

1it:
Some build troubles related.
Add "Clean previously compiled files" function.


Vitallium:
Sorry, but we don't use `build.sh` anymore. This file will be deleted soon. We have a Python script now.

PR #13721, opened 2015-11-09
Add linux build automation

mtparet:
### Description
- Add Dockerfile / circleci to build phantomjs
  This will permit to get a linux binary and verify the compilation is working.

We can download the binary for each build https://circleci.com/gh/ifeelgoods/phantomjs/3#artifacts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM ubuntu:14.04
+MAINTAINER Matthieu Paret <matthieu@ifeelgoods.com>
+# sudo docker run -v /home/matt/dev/phantomjs:/home/app 8c8a6dc2fd33 ./build.sh --confirm 
+
+RUN apt-get update
+RUN apt-get install -y build-essential g++ flex bison gperf ruby perl libsqlite3-dev libfontconfig1-dev libicu-dev libfreetype6 libssl-dev libc-dev libpng-dev libjpeg-dev python libx11-dev libxext-dev
+
+RUN mkdir -p /home/app
+ENV HOME /home/app
+WORKDIR /home/app
+
+CMD ["./build.sh"]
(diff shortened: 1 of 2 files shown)


Vitallium:
I wonder if CircleCI can handle our build. I see that they offer `Build Minutes per Month 1,500`. I don't think this is enough for PhantomJS.

mtparet:
Indeed this would lead to only 16 builds possible by month...

Vitallium:
Can they be triggered manually? And can we do builds for all supported platforms?

mtparet:
To build for windows, we need dedicated CI like http://www.appveyor.com/ for OS X, travis could still handle it. 
Here, the interesting thing is just the Dockerfile so anyone can build the image for linux.
I'm looking in travis if we could keep the binary available to download and build for Mac OS too.

Vitallium:
For Windows we have a VM in Azure. But I really want to have a service that can handle all platforms.

ariya:
@mtparet Do you mind bringing this discussion to [phantomjs-dev mailing-list](https://groups.google.com/forum/#!forum/phantomjs-dev). We've been looking for a proper CI system for PhantomJS and so far any off-the-shelf systems did not fulfill our requirements.

zackw:
I'm also working on a Docker-based build but it doesn't quite work yet, see [zackw/phantomjs/docker](https://github.com/zackw/phantomjs/tree/docker).  I'm about to go out of town for Christmas and may not have time to do anything more with it till January.

ariya:
@zackw @mtparet Let's continue the Docker-based build discussion on #13822.

ariya:
@mtparet Never mind the mailing-list, let's continue the CI build system on #13828.

PR #13722, opened 2015-11-09
fix MAKEFLAGS_JOBS parsing regex

rindeal:


Vitallium:
We have a Python script now ti build PhantomJS. Check out the master branch. Thanks!

Issue #13724, opened 2015-11-10
git commands in build docs do not auto-clone submodules


codeactual:
My PR will add add that behavior. The idea is that we'll trade the extra clone time/storage for making the repo ready to build right out of the box, avoiding surprises like `./preconfig.sh: line 187: /tmp/phantomjs/src/qt/qtbase/configure: No such file or directory` (during the Linux build) for first-time builders.

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/76a71b81f1cf7d13d43b7eabb07eccbe96dde167

PR #13725, opened 2015-11-10
Update example git commands to clone submodules

codeactual:
An attempt to address https://github.com/ariya/phantomjs/issues/13724


Vitallium:
Landed! Thanks!

PR #13726, opened 2015-11-10
add note to Mac install instructions

straz:
Inform users that the zip binary may not work on Yosemite due to a `Killed: 9` error.
Workaround exists: download and install `upx`, use it to unpack the phantomjs binary.


Vitallium:
Sorry, but this note will not be needed with the next Qt update. Qt 5.5 works well with upx.
Thanks!

PR #13755, opened 2015-11-25
[jasmine2] Avoid exception when no test was ran

madprog:
When all tests were skipped, or when no tests were found, run-jasmine2.js would raise an exception because ".alert > .bar.passed" doesn't exist.
As ".alert > .bar.skipped" exists instead, and has similar interesting content, let's use it!


zackw:
Merged.  Thanks!

PR #13766, opened 2015-11-29
#13638, Incorrect expiration dates

tmallery:
CookieJar.cookiesToMap() propagated expiration dates onto following cookies due to not clearing out the value map between cookies.

Test file: test/module/cookiejar/to-map.js

https://github.com/ariya/phantomjs/issues/13638


zackw:
Merged #13766 after updating the test case to the new test harness.

Issue #13771, opened 2015-11-30
Add ability to specify image size on output image


nwolfe-xx:
It would be great to have the option to specify the height and width of the rendered image file.  The common use-case of using PhantomJS for generating thumbnails is sometimes handled with a separate call to imageMagick to resize the image, but it would be great to do it all in one call with a minimum of chatter and temp files.  It seems like the zoomFactor option should work for this, but there seems to be a problem with that.  Also, I would rather specify the output size in pixels rather than making it a percentage of the displayPort size.

PR #13772, opened 2015-12-01
Add ability to specify the size of the output image file when rendering page

gs-nwolfe:
Usage: maxHeight and maxWidth values are set in the page.render()
options.  Aspect ratio is preserved.  If only one is specified, the
image is constrained by that dimension.  If both are specified, the
image will fit in the resulting size.

https://github.com/ariya/phantomjs/issues/13771

### DIFF
--- a/src/webpage.cpp
+++ b/src/webpage.cpp
@@ -989,6 +989,18 @@ bool WebPage::render(const QString& fileName, const QVariantMap& option)
             f = format.toLocal8Bit().constData();
         }
 
+        if( option.contains("maxWidth") && option.contains("maxHeight") ){
+            int maxWidth = option.value("maxWidth").toInt();
+            int maxHeight = option.value("maxHeight").toInt();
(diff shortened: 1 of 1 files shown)


zackw:
I like this idea, but I need to ask you for some changes.

 * Please use the same coding style as we do.  You've almost got it right, but we write our if statements like this

        if (condition) {
            // ...
        else if (condition) {
            // ...
        }

    instead of like this


        if( condition ){
            // ...
        else if( condition ){
            // ...
        }

    (The management would like to remind you that when you put spaces on the inside of your parentheses you make the baby Jesus cry.)

 * Please hook this up to work with both `render` and `renderBase64`.  To do this you will need to push the meat of your change down into `renderImage` and then arrange for `renderBase64` to take an options dictionary instead of its existing format argument.  Make sure to preserve API compatibility for the case where existing scripts pass a string.

* If at all possible, please also make it work when rendering a PDF.  It probably makes more sense for the arguments to be floating-point inches in that case.

* Please write automated tests (see `test/module/webpage/render.js`) and documentation.

* Please make sure to document how this interacts with setting the page viewport.

PR #13784, opened 2015-12-03
2.0

prasadbhima:


zackw:
As best I can tell, all of these patches are already on the development trunk.  Please clarify what you wanted us to do with them.

PR #13794, opened 2015-12-08
explain default for ssl-certificates-path better

agreif:
make clear that the default of ssl-certificates-path is the env var SSL_CERT_DIR


Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/fcbc3fb3dca9c4b66bc4aff61a7583bc2240820e Thank you!

PR #13809, opened 2015-12-15
Update Linux binary message in docs

avindra:
Since no one is working on the issue (and no progress has been made for almost a year now), it is currently misleading.


zackw:
We are, in fact, working on this.  2.1 is due Real Soon Now (probably late January) and will have official Linux binaries.

PR #13825, opened 2015-12-22
Local storage path fix

PatrickHuetter:
Integrated the localStoragePath bugfix into phantomjs 1.9.x.


zackw:
Sorry, 1.9.x development is closed.  We might take a patch for a catastrophic security hole, but that's it.

PR #13834, opened 2015-12-27
Fix bodySize in onResourceReceived end, fix #10156

### LINKED ISSUE #10156: Get amount of transfered bytes

erikdubbelboer:


erikdubbelboer:
If this gets merged #13195 can be closed.

zackw:
I don't understand the core code changes well enough to review them, but, for the test case: please try to use the Python webserver for your test case, not the PJS-embedded webserver.  It can be extended with `.py` files in the `test/www` directory.  If you can't make it work we can discuss changing it.

(The embedded webserver has architectural problems that can cause deadlocks when used like you're doing.)

erikdubbelboer:
I updated the pull request so the test uses the python webserver.

tiwilliam:
I can confirm that this PR fixes my issue in phantomas when running trying calculate size of chunked content. Would love to see this merged.

macbre/phantomas#614

abozhilov:
@erikdubbelboer any chances for new PR compatible with 2.1.1? https://github.com/ariya/phantomjs/issues/13908 breaks the current patch, as you know.

PR #13869, opened 2016-01-08
Give appropriate URL to evaluated JS files.

iradul:
This will enable remote debugger to correctly load and display available runtime scripts.
Modules will be sourced from `JAVASCRIPT_SOURCE_PLATFORM_URL` template.
Program scripts which includes main script and any script loaded using `phantom.injectJs()` will be sourced from `JAVASCRIPT_SOURCE_CODE_URL` template.

Issue #12864

### DIFF
--- a/src/phantom.cpp
+++ b/src/phantom.cpp
@@ -110,7 +110,7 @@ void Phantom::init()
         QWebSettings::setOfflineStorageDefaultQuota(m_config.offlineStorageDefaultQuota());
     }
 
-    m_page = new WebPage(this, QUrl(QString(JAVASCRIPT_SOURCE_CODE_URL).arg("html")));
+    m_page = new WebPage(this, QUrl::fromLocalFile(m_config.scriptFile()));
     m_page->setCookieJar(m_defaultCookieJar);
     m_pages.append(m_page);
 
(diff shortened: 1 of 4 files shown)


src/phantom.cpp
```diff
@@ -110,7 +110,7 @@ void Phantom::init()
         QWebSettings::setOfflineStorageDefaultQuota(m_config.offlineStorageDefaultQuota());
     }
 
-    m_page = new WebPage(this, QUrl::fromLocalFile(m_config.scriptFile()));
+    m_page = new WebPage(this, QUrl(QString(JAVASCRIPT_SOURCE_CODE_URL).arg("html")));
```
Vitallium: This will break relative paths on Windows.

iradul: This is not crucial for fixing debugger so I can revert it back to what it originally was.
I put this change so that we could have nicer and more sense-able arrangement of source files.

Vitallium:
@iradul I think you messed up the pull request a bit :-) I see wrong commits here.

iradul:
Oops I have no idea what that was ... Looks like it's fixed after rebase

Vitallium:
Landed in master! Thank you!

PR #13872, opened 2016-01-09
Improvement suggested on OSX installation doc

pechitook:
This is a suggestion to include moving the executable bin file to /usr/bin, as 3rd party libraries will try to execute it directly.

### DIFF
--- a/_posts/documentation/get-started/2000-01-01-download.md
+++ b/_posts/documentation/get-started/2000-01-01-download.md
@@ -20,7 +20,7 @@ The executable `phantomjs.exe` is ready to use.
 
 ## Mac OS X
 
-Download [phantomjs-2.0.0-macosx.zip](https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-2.0.0-macosx.zip) (13.7 MB) and extract (unzip) the content.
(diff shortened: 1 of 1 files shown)


ariya:
I think we should leave out any opinionated or recommended way to discover the executable. For example, I know a lot of advanced OS X users who don't want to tamper the system-wide directories with third-party executables (putting them in ~/bin).

p4bloch:
Sure, it is a rather opinionated way to make it available. But I also think it could help a lot to beginners. 

Advanced users can simply ignore the advice and place the bin wherever they want, it just feels to me that those who don't know how to make that bin available system wide could need further instructions.

Maybe a less opinionated way? A message telling to add the bin to their path?

ariya:
While this is true, how is this specific to PhantomJS anyway? Anytime someone downloads an executable and wants to make it available system wide will follow the same usual steps. There must be some guide on "Using command-line tools on OS X" or something similar that we can recommend.

jfeltesse-mdsol:
Just my 2 cents but 
- the guide really should mention to just use Homebrew, that's the defacto standard to install stuff on Mac OS X nowadays and [the formula is well maintained](https://github.com/Homebrew/homebrew/commits/master/Library/Formula/phantomjs.rb). Maybe mention the `--HEAD` option for the latest ref if needs be.
- +1 for **not** telling the users to "just sudo mv to /usr/bin", it's adding to the confusion since it's not explaining in any way why there and not for instance in `~/bin` or in `/usr/local/bin` (which wouldn't require you to use sudo)

ariya:
While Homebrew is nice (and I use it a lot), we try not to mention things we hardly control. In the past I did that and we always end up explaining why the version was not updated etc.

ariya:
We will leave the installation steps as is for now.

PR #13873, opened 2016-01-09
Use homebrew to install upx on OSX

mikemaccana:
Since upx website doesn't provide binaries


ariya:
This will be overtaken by event. Next 2.1 won't need UPX anymore.

mikemaccana:
Thanks! Does 'next 2.1' mean PhantomJS 2.1?

PR #13885, opened 2016-01-12
Update 2000-01-01-download.md

Lothilius:
Include quick note on how to get upx on Mac OS X.


ariya:
This will be overtaken by event. Next 2.1 won't need UPX anymore.

PR #13914, opened 2016-01-20
Google Code is dead, Mongoose is now on GitHub

mikemaccana:


Vitallium:
Landed! Thank you!

jefleponot:
Hi @Vitallium 

Do you think it could be able to compile phantomJSc with a recent version of mongoose ?
it could be great to use webSockets server connection to have a remote personal web access browser :-) or visual render like noVNC or ...

zackw:
@jefleponot The major obstacle to this is that newer versions of Mongoose are under the GPL rather than a permissive license.  Therefore, we can't upgrade without relicensing PhantomJS.

I personally would have no objection to relicensing PhantomJS core under the GPL with a note to the effect that the license does not taint programs written to the PJS API.  (It seems to me that this would make little practical difference since WebKit and Qt are already LGPL and "PhantomJS core" is not a whole lot of additional code.)  However, it is not my decision to make.

jefleponot:
Hi @zackw,

Thanks you for reply. I didn't know this problem of licence so I totally agree with you.
Do you think contributors could accept to replace Mongoose by a pure Qt web server Library that does the same things as the lastest version of mongoose ?

zackw:
> Do you think contributors could accept to replace Mongoose by a pure Qt web server Library that does the same things as the lastest version of mongoose ?

We would certainly consider a pull request that did that.  One of the biggest technical problems we have with _any_ version of mongoose is that it doesn't integrate well with the Qt event loop — a Qt-based server would make problem just go away.

We aren't going to be implementing such a server from scratch, though, and I am not aware of any existing library.  You should also be aware that one of the possible solutions to the "Webkit no longer supports Qt" problem involves not using Qt at all anymore.

Issue #13882, opened 2016-01-11
Problems reported by test/run-tests.py on OS X, Ubuntu


mepard:
I grabbed the current master and built the debug version on OS X 10.11.2 with Xcode 7.2. Then I ran test/run-tests.py. 

./build.py --confirm --debug --git-clean-qtbase --git-clean-qtwebkit > buildout.txt 2>&1
test/run-tests.py >test/testout-debug.txt 2>&1

The result is a bunch of failures due to leak reports written to stderr as well as a crash due to a failed assertion. There's also an HTTP server error. Attached are the outputs of build.py and run-tests.py as well as the crash report.

I don't know if this is new behavior as I haven't tried run-tests.py previously. 

[buildout.txt](https://github.com/ariya/phantomjs/files/85992/buildout.txt)
[testout-debug.txt](https://github.com/ariya/phantomjs/files/85993/testout-debug.txt)
[phantomjs_2016-01-11-094955_Marcs-MacBook-Pro.crash.zip](https://github.com/ariya/phantomjs/files/86002/phantomjs_2016-01-11-094955_Marcs-MacBook-Pro.crash.zip)

mepard:
I get similar results with Ubuntu-64 14.04.3.

[testout-ubuntu.txt](https://github.com/ariya/phantomjs/files/87662/testout-ubuntu.txt)

Vitallium:
Landed in master! Thank you!

PR #13915, opened 2016-01-20
Prevent test assertion

mepard:
Prevent assertion running test/module/webpage/subwindows.js. See #13882

### DIFF
--- a/src/phantom.cpp
+++ b/src/phantom.cpp
@@ -522,7 +522,15 @@ void Phantom::doExit(int code)
     emit aboutToExit(code);
     m_terminated = true;
     m_returnValue = code;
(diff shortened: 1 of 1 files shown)


mepard:
I'll try this again without the extra merges and such.

Vitallium:
FYI: You don't need to close the PR. Just update your branch and the pull request will be updated automatically. With this approach you don't need to open another pull request.

mepard:
Thanks. What I wanted to do was get rid of the extra commits, which were just to keep up with ariya:master. I'm still learning rebase, etc.

mepard:
Okay. Finally figured out how to remove unwanted commits.

Vitallium:
Landed in master! Thank you!

Issue #13121, opened 2015-04-02
Unreliable `phantom.exit` behaviours: fail to exit immediately and wrong exit status number


marco-nicola:
When I run the following script

``` javascript
phantom.exit();
for(;;) { }
```

I would expect that it ended immediately, but it doesn't: exit statement has no effects and the infinite loop predominates.

A probably related issue regards a script exit status. The execution of the following code

``` javascript
phantom.exit(1);
phantom.exit(0);
```

ends with exit status 0 (1 was expected).

Apparently a `phantom.exit` call doesn't cause an immediate exit, but it waits for _something_; the strange behaviour is noticeable adding little changes to the codes above. In fact, for example, the script

``` javascript
phantom.exit();
console.log('42');
for(;;) { }
```

exits immediately (same thing with `console.log` inside the loop) and similarly the execution of

``` javascript
phantom.exit(1);
console.log('42');
phantom.exit(0);
```

has the expected exit status 1.

Issues verified on:
- Mac OS X Yosemite 10.10.2 - PhantomJS 2.0.0 (precompiled)
- Ubuntu Server 14.04.2 LTS - PhantomJS 2.0.0 (compiled from sources)

chrisbainbridge:
Confirmed with latest git. Full example:

```                                           
var webpage = require('webpage');                                                      
var page = webpage.create();                                                           
                                                                                       
function on_page_open(status)                                                          
{                                                                                       
        if (true) {                                                  
                console.log('exit 0');                                                 
                phantom.exit(0);                                                       
        }                                                                              
        phantom.exit(1);                                                               
}                                                                                      
                                                                                       
var url = 'about:blank';                                                               
page.open(url, on_page_open); 
```
```
$ phantomjs test.js
exit 0
$ echo $?
1
```

The docs for phantom.exit should at least mention this unexpected behaviour.

PR #13917, opened 2016-01-20
Fix most leak reports from run-tests.py

mepard:
See https://github.com/ariya/phantomjs/issues/13882

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -35,6 +35,7 @@
 #include <QApplication>
 #include <QSslSocket>
 #include <QIcon>
+#include <QtWebKitWidgets/QWebPage>
 
 #include <exception>
 #include <stdio.h>
@@ -71,6 +72,13 @@ static int inner_main(int argc, char** argv)
     // execution return value
     int retVal = phantom->returnValue();
     delete phantom;
+
+#ifndef NDEBUG
+    // Prevent a bunch of leak reports that cause errors in test/run-tests.py.
+    // The leaks are only reported when compiled for debug.
+    QWebSettings::clearMemoryCaches();
+#endif
+
     return retVal;
 }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -35,7 +35,7 @@
 #include <QApplication>
 #include <QSslSocket>
 #include <QIcon>
-#include <QtWebKitWidgets/QWebPage>
+#include <QWebSettings>
 
 #include <exception>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -35,6 +35,7 @@
 #include <QApplication>
 #include <QSslSocket>
 #include <QIcon>
+#include <QtWebKitWidgets/QWebPage>
```
Vitallium: You should include `QWebSettings` instead of `QWebPage`

src/main.cpp
```diff
@@ -71,6 +72,13 @@ static int inner_main(int argc, char** argv)
     // execution return value
     int retVal = phantom->returnValue();
     delete phantom;
+
+#ifndef NDEBUG
```
Vitallium: Why not `#ifdef _DEBUG`?

zackw:
In addition to Vitallium's line notes, I want to bring issue #13121 to your attention - to fix this we're probably going to have to make phantom.exit() call the C library's exit(), which means app.exec() will never return.  We don't want to lose leak detection in debug builds by doing that, but it may take some finagling.

src/main.cpp
zackw: `_DEBUG` isn't defined.  I think what we actually want here is `#ifndef QT_NO_DEBUG`.

Vitallium: ah, yes, `_DEBUG` is defined by MSVC only. So yes, agreed on `QT_NO_DEBUG`

mepard: How about checking QT_DEBUG? I've never liked negatively expressed variables. WTF used NDEBUG where it reported the leak, so that's what I used here.

src/main.cpp
mepard: Duh. I had done some other experiments and this was left over.

src/main.cpp
zackw: I don't like them either, but unfortunately, there is no positive QT_DEBUG, only QT_NO_DEBUG in release builds.

mepard:
That'll be tricky because Phantom needs to be destructed in addition to calling clearMemoryCaches or there will still be leaks.

zackw:
I've gone ahead and landed this with the changes as discussed.  There are two remaining testsuite failures introduced by debug mode:

```
module/webpage/includejs: ERROR
ERROR: Unexpected output on stderr
  ASSERTION FAILED: !m_frame->page()->defersLoading() || InspectorInstrumentation::isDebuggerPaused(m_frame)
  loader/DocumentLoader.cpp(361) : void WebCore::DocumentLoader::finishedLoading(double)
  1   0x1fe6292 /home/zack/projects/phantomjs/bin/phantomjs() [0x1fe6292]
[...]
ERROR: PhantomJS killed by signal 11

regression/pjs-10690: ERROR
ERROR: Unexpected output on stderr
  LEAK: 1 CachedResource
```

The leak is your issue #13920.  I've filed #13935 for the assertion.

zackw:
> That'll be tricky because Phantom needs to be destructed in addition to calling clearMemoryCaches or there will still be leaks.

Yeah, I know. It might end up having to do something pretty nasty.

Issue #12903, opened 2015-01-25
Memory Leak: Phantomjs 2 checked out on 13th November 2014


surenderthakran:
Phantomjs version: Phantomjs 2 (built from source on 13th Nov 14 from master)

I wrote a script to scan websites using phantomjs. I open not more than 4 simultaneous pages at any point.
Some serious memory leak happens when i run it 

> page.settings.loadImages = false;

I run my script on an amazon linux ec2 server with a 1GB ram and a single core processor (its one of the lighter servers).

When not loading images my ram usage gradually increases to 99% as seen by the "top" command before phantom crashes.
However, when loading images it stabilizes at around 6-7%.

alexey89777:
+1 have the same issue too

ariya:
Can you build the tag `2.0.0` and try it again? If that still happens, I'll mark this as Regression.

surenderthakran:
@ariya ok will try again on 2.0.0

iradul:
+1

iradul:
This is WebKit related issue :
https://bugreports.qt.io/browse/QTBUG-34494 (https://codereview.qt-project.org/#/c/76934/)
https://bugreports.qt.io/browse/QTBUG-36530
https://bugreports.qt.io/browse/QTBUG-38857

Take a look at this test script, you'll see how memory usage grows up to 1GB with ```--load-images=false``` but with ```--load-images=true``` it doesn't grow more then 350MB:
```javascript
var page = require('webpage').create(),
[...]

PR #13919, opened 2016-01-20
Test for leaks when --load-images=no.

mepard:
This test, when run with phantomjs built with --debug, demonstrates the leak described in #12903. The html loads specifies two images and results in the output below. I've varied the number of images and it appears to leak the number of images + 1 of each of the two objects. I'm investigating.

module/webpage/load-images-no: ERROR
ERROR: Unexpected output on stderr
  LEAK: 3 CachedResource
  LEAK: 3 WebCoreNode

### DIFF
--- a/test/module/webpage/load-images-no.js
+++ b/test/module/webpage/load-images-no.js
@@ -1,0 +1,16 @@
+//! phantomjs: --load-images=no
+
+// Compiled for debug, this tests for leaks when not loading images.
+//
+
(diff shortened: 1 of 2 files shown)


zackw:
Could you please augment the test so that it also tests for images genuinely not getting loaded in `--load-images=no` mode?  Then it would be doing something useful in both debug and release builds.

mepard:
Replacing this PR with #13980.

PR #13922, opened 2016-01-21
Remove capturing response feature

vitallium:
Issue: #13908

/cc @ariya @zackw


erikdubbelboer:
Is there no other way in which we can bring back this capturing feature? Is there maybe some signal missing which makes synchronous AJAX fail. Maybe something networkreplyproxy needs to implement but doesn't.

ariya:
:+1:

wi-ski:
How is this still a thing. :(

PR #13925, opened 2016-01-22
Update 2000-01-01-download.md

yingtho:
Added an easier install on macos.


ariya:
This has been discussed. The first priority is always any download from the the official PhantomJS team.

PR #13945, opened 2016-01-26
Added stdin with a close() method to child process

tdammers:
Based on https://github.com/execjosh/phantomjs/commit/a00d70c5b61aef475cd9bc8b374e1113e21ba59e, but additionally supports `stdin.close()`, which is required for using subprocesses that wait for EOF on stdin before exiting.

### DIFF
--- a/test/module/child_process/basics.js
+++ b/test/module/child_process/basics.js
@@ -1,10 +1,34 @@
 // Test basic child process functionality
 //
 
+var ECHO_SCRIPT = 'echo.py';
+var CAT_SCRIPT = 'cat.py';
+
+setup(function(){
+    var f;
+    var fs = require('fs');
+
+    // Platform-agnostic drop-in replacements for 'echo -n' and 'cat'.
+    // Using sys.stdout.write rather than print here to make the
+    // python code 2/3 agnostic.
+
+    f = fs.open(ECHO_SCRIPT, 'w');
+    f.writeLine('import sys');
+    f.writeLine('');
+    f.writeLine('sys.stdout.write(" ".join(sys.argv[1:]))');
+    f.close();
+
+    f = fs.open(CAT_SCRIPT, 'w');
+    f.writeLine('import sys');
+    f.writeLine('');
+    f.writeLine('sys.stdout.write(sys.stdin.read())');
+    f.close();
+});
+
 async_test(function(){
     var text = "hello";
     var process = require('child_process');
-    var p = process.spawn("echo", ["-n", text]);
+    var p = process.spawn("python", [ECHO_SCRIPT, text]);
     var out = '';
     p.stdout.on('data', function(data) { out += data; });
     p.on('exit', this.step_func_done(
@@ -17,7 +41,7 @@ async_test(function(){
 async_test(function(){
     var text = "hello";
     var process = require('child_process');
-    var p = process.spawn("cat", []);
+    var p = process.spawn("python", [CAT_SCRIPT]);
     var out = '';
     p.stdout.on('data', function(data) { out += data; });
     p.on('exit', this.step_func_done(
--- a/test/module/child_process/basics.js
+++ b/test/module/child_process/basics.js
@@ -4,7 +4,7 @@
 var ECHO_SCRIPT = 'echo.py';
 var CAT_SCRIPT = 'cat.py';
 
-setup(function(){
+var createScripts = function(){
     var f;
     var fs = require('fs');
 
@@ -23,9 +23,17 @@ setup(function(){
     f.writeLine('');
     f.writeLine('sys.stdout.write(sys.stdin.read())');
     f.close();
-});
+};
+
+var deleteScripts = function() {
+    var fs = require('fs');
+    fs.remove(CAT_SCRIPT);
+    fs.remove(ECHO_SCRIPT);
+};
 
 async_test(function(){
+    createScripts();
+    this.add_cleanup(deleteScripts);
     var text = "hello";
     var process = require('child_process');
     var p = process.spawn("python", [ECHO_SCRIPT, text]);
@@ -39,6 +47,8 @@ async_test(function(){
 }, "call a simple subprocess");
 
 async_test(function(){
+    createScripts();
+    this.add_cleanup(deleteScripts);
     var text = "hello";
     var process = require('child_process');
     var p = process.spawn("python", [CAT_SCRIPT]);
--- a/test/module/child_process/basics.js
+++ b/test/module/child_process/basics.js
@@ -1,0 +1,30 @@
+// Test basic child process functionality
+//
+
+async_test(function(){
+    var text = "hello";
+    var process = require('child_process');
+    var p = process.spawn("echo", ["-n", text]);
+    var out = '';
+    p.stdout.on('data', function(data) { out += data; });
+    p.on('exit', this.step_func_done(
+                    function(exitCode) {
+                        assert_equals(exitCode, 0);
+                        assert_equals(out, text);
+                    }));
+}, "call a simple subprocess");
+
+async_test(function(){
+    var text = "hello";
+    var process = require('child_process');
+    var p = process.spawn("cat", []);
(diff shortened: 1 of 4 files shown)


zackw:
This looks good as far as it goes, but I would like to ask you to do a little bit more work before we merge it:

* Please write some test cases.  I think the subprocess module currently isn't being tested at all.
* Please add a `_write` overload that takes a `QByteArray`, so that *binary* data can be written to a subprocess's stdin.  (I *think* this is all you have to do to expose that on the JS side - Qt will just magically map a JS typedarray into a QByteArray and call the appropriate function.)
* Your `proc.stdin.close()` method appears to correspond to Node `proc.stdin.end()`, which is currently stubbed out.  Please either make both `close` and `end` do the same thing, or rename your `close` method to `end` and remove the stub; your call.

zackw:
And thank you for picking up the ball, here.

tdammers:
- Added some basic tests that cover a simple (pre-PR) use case for a child process, and a use case where the child process reads and echoes stdin.
- WRT the `end`/`close` functionality, I opted to make both methods do the same thing; the documented stream API has `close`, not `end`, so I figured it'd be best to stay true to that.
- Binary input to stdin is as of yet unsolved; the code there isn't mine, so I'd have to dig in a bit deeper, and I'm a bit short on time right now, but maybe I can take a stab at it tomorrow.

tdammers:
On further investigation, it seems that child process stdout and stderr do not support raw output either, so I believe supporting raw binary I/O on child processes is best left to a separate ticket, particularly because the API isn't obvious.

Also, JavaScript strings are not the right data type to pass the raw binary data in, as they are sequences of characters, not sequences of bytes, and I'm not exactly sure what happens encoding-wise if you marshal JS strings to QByteArrays directly.

zackw:
> On further investigation, it seems that child process stdout and stderr do not support raw output either, so I believe supporting raw binary I/O on child processes is best left to a separate ticket, particularly because the API isn't obvious.

OK, that's fair.  Do you want to file another ticket or shall I?

> Also, JavaScript strings are not the right data type to pass the raw binary data in

Yes.  The desired JS-side API would be for write() to accept [typed arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays), which would then show up on the C++ side as [QByteArray](http://doc.qt.io/qt-5/qbytearray.html) objects (if I understand correctly, the JS-to-QT bridge does this automatically).  On the read side, I'm not sure what the right API is - the JavaScript code's expectations would need to be specified somehow.  Do you know what Node does?

tdammers:
> Yes. The desired JS-side API would be for write() to accept typed arrays, which would then show up on the C++ side as QByteArray objects (if I understand correctly, the JS-to-QT bridge does this automatically). On the read side, I'm not sure what the right API is - the JavaScript code's expectations would need to be specified somehow. Do you know what Node does?

Frankly, I've never ventured this deeply into JavaScript/C++ interop before, so I don't really know. A quick node.js API search reveals this: https://nodejs.org/api/stream.html#stream_new_stream_readable_options, which suggests that streams receive an encoding property at creation time, and if that is set to `null`, `read` will return the raw buffer rather than a string; the `write` API however simply dispatches on the type of its first argument, which must be either a string (in which case it encodes) or a buffer (in which case it writes raw bytes). The Node API documentation does not specify which mode a child process' stdout and stderr are in however, and doesn't document any ways of putting them into raw mode. There's `setEncoding`, and I'd assume that passing `null` as the encoding to that would cause the stream to go into raw mode, but the documentation isn't explicit about this.

zackw:
Filed #13952 on the future support for binary I/O.

test/module/child_process/basics.js
```diff
@@ -0,0 +1,30 @@
+// Test basic child process functionality
+//
+
+async_test(function(){
+    var text = "hello";
+    var process = require('child_process');
+    var p = process.spawn("echo", ["-n", text]);
```
zackw: @Vitallium Can you comment on whether this is safe on Windows?

test/module/child_process/basics.js
```diff
@@ -0,0 +1,30 @@
...
+                    function(exitCode) {
+                        assert_equals(exitCode, 0);
+                        assert_equals(out, text);
+                    }));
+}, "call a simple subprocess");
+
+async_test(function(){
+    var text = "hello";
+    var process = require('child_process');
+    var p = process.spawn("cat", []);
```
zackw: @Vitallium Here also.

test/module/child_process/basics.js
Vitallium: Command `echo` doesn't need argument `-n` on Windows. Otherwise the output will be `-n hello`

test/module/child_process/basics.js
Vitallium: `cat` is not available on Windows.

JamesMGreene: `type con` is basically equivalent. See http://ss64.com/nt/type.html

test/module/child_process/basics.js
JamesMGreene: Closest thing I'm aware of is this: http://winaero.com/blog/how-to-make-the-echo-command-without-new-line-in-windows/

JamesMGreene: ****cough**** Seems like our life would be so much easier with Chromium + Node.js. :wink:

test/module/child_process/basics.js
tdammers: Is there an easy way to branch by OS in the tests?

L09:
Can we also pass the exit code? In child_process.js, line 91:
```
ctx.on("exit", function (code) {
    return cb(null, stdout, stderr)
  })
```
should be replaced with the following lines:
```
ctx.on("exit", function (code) {
    return cb(code === 0 ? null : {'code' : code}, stdout, stderr)
  })
```

tdammers:
@L09: I'd say that sounds like a reasonable idea, but I don't think it's directly related to this PR, best to keep it a separate issue.

L09:
Thanks. It's a one-line change, so IMHO it does not deserve a separate PR. We already catch the exit code in childprocess.cpp (see the _finished and _error methods).

execjosh:
@tdammers Hey, thanks for picking this up!

tdammers:
@execjosh no problem.

Re the test cases; I was thinking that maybe the smoothest cross-platform way of testing subprocesses would be to write a little Python script and shell out to that, rather than rely on anything platform-specific to be installed? We could basically write Python drop-in replacements for `cat` and `echo`, methinks.

tdammers:
OK, this should work on any platform that has Python installed.

@zackw does this look like it could be merged?

zackw:
@tdammers I was about to merge this, but, um, you forgot to check in echo.py and cat.py :-(

zackw:
@L09 I looked at your 'one-line change' and it is actually the tip of an iceberg, so I've split it out to #13970. If you have time to tackle the audit I mentioned there, that would be appreciated.

zackw:
Merged with small modifications.  The most significant is: If you're running the test suite at all, we know you have Python *somewhere* (since the test runner itself is written in Python) but it might not be in $PATH.  Therefore, I made `run-tests.py` pass down the value of `sys.executable` to all tests, and the child process tests use that as the Python interpreter rather than just "python".  Also, `echo.py` and `cat.py` now live under test/lib/fixtures rather than being embedded in the test itself.

(and somehow this involved fixing like five unrelated bugs in the test suite :-P)

ariya:
:+1: @tdammers

tdammers:
@zackw great! I didn't realize there was a better fixtures mechanism in place, this is a much saner approach.

BTW., would be nice if someone could verify that the tests run fine on Windows, because I don't have a Windows machine around to test on.

mepard:
This test passes on Windows 10.

    module\child_process\basics: pass

Others are failing. See #13974

execjosh:
Merged!  Yay!

PR #13971, opened 2016-02-04
Linting the javascript code

LinusU:
As discussed in #13314 I have gone thru the code with ESLint to make it follow Standard style. There is one commit per linting rule as requested.

Reviewing this is probably easiest to do commit by commit.

ping @zackw 

[js-standard-style](https://github.com/feross/standard)

### DIFF
--- a/src/modules/child_process.js
+++ b/src/modules/child_process.js
@@ -175,18 +175,6 @@ function newContext () {
   return ctx
 }
 
-function delayCallback () {
-  var args = arguments.length > 0 ? [].slice.call(arguments, 0) : []
-  var fn = args.shift()
-  if (!isFunc(fn)) {
-    return
-  }
-  var that = this
-  setTimeout(function () {
-    fn.apply(that, args)
-  }, 0)
-}
-
 function isFunction (o) {
   return typeof o === 'function'
 }
--- a/src/modules/child_process.js
+++ b/src/modules/child_process.js
@@ -42,7 +42,7 @@ exports.spawn = function (cmd, args, opts) {
     opts = {}
   }
 
-  opts.encoding = opts.encoding || "utf8"
+  opts.encoding = opts.encoding || 'utf8'
   ctx._setEncoding(opts.encoding)
 
   ctx._start(cmd, args)
@@ -58,7 +58,7 @@ exports.exec = function (cmd, opts, cb) {
     cb = NOP
   }
 
-  return cb(new Error("NotYetImplemented"))
+  return cb(new Error('NotYetImplemented'))
 }
 
 /**
@@ -75,20 +75,20 @@ exports.execFile = function (cmd, args, opts, cb) {
     opts = {}
   }
 
-  opts.encoding = opts.encoding || "utf8"
+  opts.encoding = opts.encoding || 'utf8'
   ctx._setEncoding(opts.encoding)
 
-  var stdout = ""
-  ctx.stdout.on("data", function (chunk) {
+  var stdout = ''
+  ctx.stdout.on('data', function (chunk) {
     stdout += chunk
   })
 
-  var stderr = ""
-  ctx.stderr.on("data", function (chunk) {
+  var stderr = ''
+  ctx.stderr.on('data', function (chunk) {
     stderr += chunk
   })
 
-  ctx.on("exit", function (code) {
+  ctx.on('exit', function (code) {
     return cb(null, stdout, stderr)
   })
 
@@ -101,7 +101,7 @@ exports.execFile = function (cmd, args, opts, cb) {
  * fork(modulePath, [args], [options])
  */
 exports.fork = function (modulePath, args, opts) {
-  throw new Error("NotYetImplemented")
+  throw new Error('NotYetImplemented')
 }
 
 // private
@@ -115,7 +115,7 @@ function newContext () {
     var handler
 
     switch (evt) {
-      case "exit":
+      case 'exit':
         handler = ctx[evt]
         break
       default:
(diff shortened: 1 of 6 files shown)


Vitallium:
What about using this https://github.com/Flet/semistandard ? Yes, I know about semicolons in JavaScript. But I like semicolons!

LinusU:
I personally prefer without semicolons but it would be easy to update the pull request to include them.

Should be noted that [standard](https://npmjs.com/packages/standard) is much more popular compared to [semistandard](https://www.npmjs.com/package/semistandard). `132 872` downloads in the last month versus `10 329`.

zackw:
When it comes to JavaScript, I also have a strong preference for fully semicolon-ed code.  The "automatic semicolon insertion" rules are somewhat counterintuitive, and relying on them can get you into trouble.

zackw:
So here's how I suggest we move forward on this.

I am going to annotate each commit with either :+1: or :-1:, and in the latter case, an explanation of why I don't like the changes made.  You rebase your patch series with all of the :+1: commits shuffled before all of the :-1: commits.  Since the weekend is coming up, I will give @ariya and @Vitallium until Tuesday morning (Feb. 9) to file objections, and then (assuming there are none) I will merge all of the :+1: commits.  That cuts down your merge burden and we can then iterate on the rest of it.

Do we want to think about having some degree of automatic checking for new lint errors in our JS?

zackw:
Oh, also, did you lint anything outside `src/`?

ariya:
To reduce the risk I recommend the following.

First, separate all the style-related change (cosmetics) that does **not** change the semantics of the code. This includes e.g. unifying string literals to use single quote, adding/removing whitespaces, etc. In this step, try to adhere as much possible to existing code, e.g. since we use semicolons, don't remove them. Instead of using standards, this is easier done using JSCS. Take a look at my other project, Esprima, where checking for this is a simple JSCS call (https://github.com/jquery/esprima/blob/master/package.json#L75). Side note: @mikesherov and I gave a talk on this subject: https://www.youtube.com/watch?v=954lRqSI_Ro.

After that, the semantics-related change needs to be evaluated carefully. Changing `==` to `===` is not to be taken lightly. This is not about style preference or popularity. This is simply because it is not always obvious (this is JavaScript after all) why it was written that way and thus changing it without taking a pause and think can lead to a latent defect.

I think it's easier for @zackw and me (and others) to review the change if you create a dedicated pull request for the first step first (style changes) and it's fine to have it in a series of small commits.

LinusU:
> This includes e.g. unifying string literals to use single quote

Those are semantically the same in javascript

> , adding/removing whitespaces, etc.

That shouldn't change the semantics of the code? Unless your breaking ASI which I aren't doing...

>  Changing `==` to `===` is not to be taken lightly.

True, but I went thru each change very carefully and I'm 100% that they don't change anything. I've commented why on that commit as well...

> Oh, also, did you lint anything outside `src/`?

Nope :)

> So here's how I suggest we move forward on this.
> 
> ...

That sounds good :+1: 

> Do we want to think about having some degree of automatic checking for new lint errors in our JS?

The plan here was to be able to use the `standard` command (which is eslint with a provided config) but if you want to stick to 4 spaces that won't work (semicolons could be solved by using semistandard). One way would be to write our own eslint-config.

ariya:
> This includes e.g. unifying string literals to use single quote
> Those are semantically the same in javascript
> 
> , adding/removing whitespaces, etc.
> That shouldn't change the semantics of the code? Unless your breaking ASI which I aren't doing...

Not sure I follow. I completely agree that this doesn't change semantics as that was exactly what I wrote.

> The plan here was to be able to use the `standard` command...

`standard` is simply too opinionated. I agree with @zackw that we should pick a style as close as possible to the existing style and other surrounding code.

@LinusU Did you take a look at JSCS?

LinusU:
> Not sure I follow. I completely agree that this doesn't change semantics as that was exactly what I wrote.

Sorry, I must have misread

> `standard` is simply too opinionated.

That's the beauty with it :)

> Did you take a look at JSCS?

Haven't looked at it too much but from what I have understand it doesn't recognise potential code bugs which eslint does. E.g. unused variables, accidental globals...

Looking at npm it seems like eslint is more popular, with more than twice the downloads. I guess it doesn't matter too much though...

ariya:
> standard is simply too opinionated.
> That's the beauty with it :)

Sure, if you want to start a new project from scratch. But here in this context, we have to consider many different factors, including a close resemblance to other related code. Our JavaScript code is minuscule compared to C++, what good it is to change all JavaScript code to indent with 2-space if the majority is 4-space C++ code?

> Did you take a look at JSCS?
> Haven't looked at it too much but from what I have understand it doesn't recognise potential code bugs which eslint does. E.g. unused variables, accidental globals...

That's precisely the point. Let's separate style unification vs potential code bugs.

> Looking at npm it seems like eslint is more popular, with more than twice the downloads. I guess it doesn't matter too much though...

We all love ESLint, but popularity here is not a factor that improves software quality.

LinusU:
> Sure, if you want to start a new project from scratch. But here in this context, we have to consider many different factors, including a close resemblance to other related code. Our JavaScript code is minuscule compared to C++, what good it is to change all JavaScript code to indent with 2-space if the majority is 4-space C++ code?

Yeah I understand where you're coming from, I just personally like standard very much but I understand why you might not want to adapt it.

How do you feel about rebasing with only the things that have been :+1:ed and then we can move forward with JSCS after that?

ariya:
> ... then we can move forward with JSCS after that?

I think we should investigate JSCS use before making that decision. Otherwise, that may lead to another additional work.

ariya:
Closing to due to lack of follow-ups.

Issue #13976, opened 2016-02-04
global variables are forbidden in strict mode, but they exists on rasterize.js


yuxel:
it seems "strict mode" enabled ~2 months ago with https://github.com/ariya/phantomjs/commit/9ac8cccf0abd7e2675d943d690fab0631b9a68bd#diff-b1bce6d16bf409f61061d23b1850eb43R1

But, we still have some global variables in rasterize.js.

Such command gives error below

<pre>$ phantomjs phantomjs/examples/rasterize.js http://google.com output.png 1300px*1000px
ReferenceError: Strict mode forbids implicit creation of global property 'pageWidth'

  phantomjs://code/rasterize.js:23 in global code</pre>

Vitallium:
Fixed in https://github.com/ariya/phantomjs/commit/0ba7bc8ef6867ff335402e9b58ae8558ca8a54d9

PR #13978, opened 2016-02-04
#13976 - 'global variables are forbidden in strict mode' issue fixed

yuxel:
moved global vars into scope with adding them to variable list


Vitallium:
Thank you! Landed in https://github.com/ariya/phantomjs/commit/0ba7bc8ef6867ff335402e9b58ae8558ca8a54d9

PR #13980, opened 2016-02-05
Test load-images=yes and load-images=no.

mepard:
Tests the function of the --load-images command line option. When compiled with --debug, demonstrates the leak described in #12903.

Replaces previous pull request #13919.

### DIFF
--- a/test/lib/www/load-images.html
+++ b/test/lib/www/load-images.html
@@ -1,6 +1,9 @@
+<!DOCTYPE html>
 <html>
+<head>
+</head>
 <body>
-    <img src="logo.png"/>
-    <img src="phantomjs.png"/>
+    <img src="logo.png" alt="logo">
+    <img src="phantomjs.png" alt="phantomjs">
 </body>
 </html>
--- a/test/lib/www/load-images.html
+++ b/test/lib/www/load-images.html
@@ -1,0 +1,6 @@
+<html>
+<body>
+    <img src="logo.png"/>
+    <img src="phantomjs.png"/>
+</body>
+</html>
--- a/test/module/webpage/load-images-no.js
+++ b/test/module/webpage/load-images-no.js
@@ -1,0 +1,52 @@
+//! phantomjs: --load-images=no
+
+// Compiled for debug, this tests for leaks when not loading images.
+//
+
+var url = TEST_HTTP_BASE + 'load-images.html';
+
+function do_test() {
+    var page = require('webpage').create();
+
+    var numImagesRequested = 0;
+    var numImagesLoaded = 0;
+    
+    page.onResourceRequested = this.step_func(function (resource) {
+        if (/\.png$/.test(resource.url)) ++numImagesRequested;
+    });
+
+    page.onResourceReceived = this.step_func(function (resource) {
+        if (resource.stage === 'end' && /\.png$/.test(resource.url)) ++numImagesLoaded;
+    });
+
+    function checkImageCounts ()
+    {
+        var imageCounts = page.evaluate (function() {    
+            "use strict";
+            try
+            {
+                var images = document.getElementsByTagName('img');
+                var imageCounts = {};
+                imageCounts.numImages = images.length;
+                imageCounts.numComplete = Array.prototype.reduce.call(images, function(n, image) { return n + (image.complete? 1 : 0); }, 0);
+                return imageCounts;
+            }
+            catch(err)
+            {
+                return undefined;
+            }
+        });
+        assert_equals(imageCounts.numImages, 2);
+        assert_equals(imageCounts.numComplete, 2);  // Images are complete even when loading suppressed.
+        assert_equals(numImagesRequested, 0);
+        assert_equals(numImagesLoaded, 0);
+        page.release();
+    }
+    
+    page.open(url, this.step_func_done (function (status) {
+        assert_equals(status, "success");
+        checkImageCounts ();
+    }));
+}
+
+async_test(do_test, "load a page two images with --load-images=no");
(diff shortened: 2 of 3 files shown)


test/lib/www/load-images.html
```diff
@@ -0,0 +1,6 @@
+<html>
+<body>
+    <img src="logo.png"/>
+    <img src="phantomjs.png"/>
```
zackw: Please don't use XMLisms (``) in a file that isn't XML.  (Please *do* add `<!doctype html>` at the top of this file.)

test/module/webpage/load-images-no.js
```diff
@@ -0,0 +1,52 @@
...
+        page.release();
+    }
+    
+    page.open(url, this.step_func_done (function (status) {
+        assert_equals(status, "success");
+        checkImageCounts ();
+    }));
+}
+
+async_test(do_test, "load a page two images with --load-images=no");
```
zackw: Is there a reason `do_test` is not an anonymous function inside the call to `async_test`?

test/module/webpage/load-images-no.js
```diff
@@ -0,0 +1,52 @@
+//! phantomjs: --load-images=no
+
+// Compiled for debug, this tests for leaks when not loading images.
+//
```
zackw: Is the line with just `//` on it an editing mistake?

test/module/webpage/load-images-no.js
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+    page.onResourceReceived = this.step_func(function (resource) {
+        if (resource.stage === 'end' && /\.png$/.test(resource.url)) ++numImagesLoaded;
+    });
+
+    function checkImageCounts ()
+    {
+        var imageCounts = page.evaluate (function() {    
+            "use strict";
```
zackw: If you're going to `"use strict"` please do it for the entire file.

test/module/webpage/load-images-no.js
```diff
@@ -0,0 +1,52 @@
...
+            {
+                var images = document.getElementsByTagName('img');
+                var imageCounts = {};
+                imageCounts.numImages = images.length;
+                imageCounts.numComplete = Array.prototype.reduce.call(images, function(n, image) { return n + (image.complete? 1 : 0); }, 0);
+                return imageCounts;
+            }
+            catch(err)
+            {
+                return undefined;
```
zackw: What exceptions might get thrown in this context?

test/module/webpage/load-images-no.js
mepard: These evolved from pjs-10690.js. Other than that, no reason at all. I personally wouldn't make it anonymous, though. I've gotten in the habit of naming all functions to improve traces (particularly in casperjs errors).

test/lib/www/load-images.html
mepard: I copied logo.html and didn't give it much thought.

test/module/webpage/load-images-no.js
mepard: Does that work when it's inside evaluate?

test/module/webpage/load-images-no.js
mepard: Don't know, but I've gotten into the habit of making sure problems in evaluate show up outside in some fashion.

test/module/webpage/load-images-no.js
zackw: Oh, good point, it probably doesn't.  I think we serialize to text and recompile in the page context.  Leave that there and add one for the file, then?

mepard: Sounds good.

test/module/webpage/load-images-no.js
mepard: Yup. Sorry.

test/module/webpage/load-images-no.js
zackw: I'm not sure what happens if an exception is thrown and not caught inside `evaluate` -- but this way, the exception is swallowed and we just have `undefined` in a comparison later, which is safe (`assert_*` use strict equality) but might be baffling.  I guess what I want is to make sure that any potential exception gets logged somehow.

test/module/webpage/load-images-no.js
zackw: I was mainly just curious; most of the tests use anon functions for this.

test/lib/www/load-images.html
zackw: Yeah that file could probably stand to have the dust blown off it a little.  No worries.

test/module/webpage/load-images-no.js
mepard: Logging inside evaluate won't surface unless you use onConsoleMessage.

I played around with what happens when I intentionally throw an exception. In this case, it's not the return value that gets checked, but returnvalue.someField. This results in e.g.

	module/webpage/load-images-yes: FAIL
	 FAIL: load a page two images with --load-images=yes
	  null is not an object (evaluating 'imageCounts.numImages')
	  checkImageCounts@phantomjs://code/load-images-yes.js:38:34
	  pageOpened@phantomjs://code/load-images-yes.js:46:25

If I remove the try/catch from evaluate, the exception does get reported, but it's pretty strange:

	module/webpage/load-images-yes: ERROR
	ERROR: neither a test nor a diagnostic
	  Error: oops
	 FAIL: test 1 did not report status

Error: oops is the exception I added, but in most cases you wouldn't see the explicit throw in the test code. The other way seems clearer to me.

mepard:
https://github.com/Vitallium/qtwebkit/pull/8

ariya:
Excellent fix! Please go ahead and squash it into one commit and I'll land it. Thanks!

Vitallium:
Squashed and landed in https://github.com/ariya/phantomjs/commit/4ef5f2760028f8f0c341bbad3915937587d93b3f

PR #13981, opened 2016-02-06
Update 2000-01-01-quick-start.md - added trailing slash to example.com/ url

justinlevi:
I was trying to go through this quickstart and wasted a bunch of time trying to figure out why the simple example wasn't working for me. I entered `google.com` because the example uses `example.com`.

``` javascript
page.open('http://google.com', function(status) {
  console.log("Status: " + status);
  if(status === "success") {
    page.render('google.png');
  }
  phantom.exit();
});
```

I finally checked the screen-capture documentation : `http://phantomjs.org/screen-capture.html` and noticed that nearly the same example adds a trailing slash to the url. 

``` javascript
page.open('http://github.com/', function() {
  page.render('github.png');
  phantom.exit();
});
```

Hopefully this saves someone else a small headache.

### DIFF
--- a/_posts/documentation/learn/2000-01-01-quick-start.md
+++ b/_posts/documentation/learn/2000-01-01-quick-start.md
@@ -44,7 +44,8 @@ The following script demonstrates the simplest use of page object. It loads exam
 
 ```javascript
 var page = require('webpage').create();
-page.open('http://example.com', function(status) {
+page.open('http://example.com/', function(status) {
+  // Trailing slash on the above url is required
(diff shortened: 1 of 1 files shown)


ariya:
The trailing backslash should not be necessary. I can't reproduce the problem at all.
Could you give a detailed info about your system etc?

justinlevi:
@ariya I have to apologize sir as I'm unable to duplicate today. I knew I should have done a screen recording the other day when I was running into this. If I see it again, I will try and grab more details.

PR #13988, opened 2016-02-09
add rpm build script

jefleponot:
Hello,

Could you add the rpm package installer to phantom 2.1?

Thanks in advance,
Jef

### DIFF
--- a/rpm/mkrpms.sh
+++ b/rpm/mkrpms.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+#
+# A silly little helper script to build the RPM.
+set -e
+
+phantomdir=$(basename $(dirname $(pwd)))
+name=${1:?"Usage: build <toolname>"}
+name=${name%.spec}
+topdir=$(mktemp -d)
+version=$(../bin/phantomjs --version | cut -d. -f1-2 )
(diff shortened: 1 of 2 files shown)


ariya:
See #13988. Those files get outdated pretty quickly and unless a maintainer steps us, we prefer not to bear an additional burden of maintaining it.

jefleponot:
@ariya, Thank you for reply 

my PR is not require maintainer because it is generic : 
- version and release are extract from "phantomjs --version" command
- files created depend on files includes in the project /usr/bin/phantomjs and /usr/share/phantomjs/*

older versions of mkrpm.sh wasn't generic and require to compile all source ! now it's light and minimalist. 

But it is up to you Boss :+1: 
jef

ariya:
It's still considered maintenance. If for some reason PhantomJS code is changed in such a way that it breaks the rpm build, we would never know it unless someone runs the build script.

I recommend creating a (separate) repository that contains a script to:
- grab PhantomJS code
- place the necessary build script
- build the rpm

You probably can run a hosted CI service that will try to build the rpm, and hence providing a safety mechanism to detect a potential build failure.

jefleponot:
You're totally right. The solution you proposed is the best one. 

Thanks a lot for all your work.

PR #14015, opened 2016-02-17
Actual implementation of custom IP in ghostdriver

iiLaurens:
Fixes a simple issue with Ghostdriver not using custom defined ip as per #12553.

The problem is solved by allowing Ghostdriver to use the given IP that is set default by the config or as given by a service argument. This allows the use of PhantomJS webdriver in the Openshift PaaS environment.


Vitallium:
Thanks for your contribution! But all Ghostdriver related fixes should be landed to its repository first: https://github.com/detro/ghostdriver

Issue #10687, opened 2012-08-01
Phantomjs does not exit if file has Parse Errors


ahawkins:
_**[m...@broadcastingadam.com](http://code.google.com/u/116377228668850173159/) commented:**_
> <b>Which version of PhantomJS are you using? Tip: run 'phantomjs --version'.</b>

> 1.6.1
> 
> <b>What steps will reproduce the problem?</b>
1. Create a file with only this line: thisIsUndefined!. Save the file as &quot;foo.js&quot;
> 2. $ phantomjs foo.js
> <b>3.</b>

> <b>What is the expected output? What do you see instead?</b>

> adam at mba : ~ % phantomjs foo.js
> SyntaxError: Parse error
> .....
> 
> phantomjs just hangs. The process should exit(1) if there is a syntax error in the executing file.

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #687](http://code.google.com/p/phantomjs/issues/detail?id=687).
:star2: &nbsp; **6** people had starred this issue at the time of migration.

cyphunk:
how I get around this. When script hangs, execute with nodejs to see if syntax error is at fault.

graceful-code:
If page.onError or window.onerror could listen for these syntax and runtime errors, then at least we could catch the errors and let Phantom move on instead of just hanging. I believe this is related to #10537

GDmac:
I bumped into this with the rasterize.js example script linked from the screen capture page
http://phantomjs.org/screen-capture.html

Testing from terminal outputs following message and does not exit (need ctrl-c)
``ReferenceError: Strict mode forbids implicit creation of global property 'pageWidth'
phantomjs://code/rasterize.js:23 in global code
``

I found another tip on stackoverflow to wrap the script in a try-catch block,
http://stackoverflow.com/questions/11180662/automatically-close-phantomjs-after-running-script

It would sure be nice to have an option to have PhantomJS just exit on parse error...

PhantomJS 2.1.1 (macosx)

szakharchenko:
@acidus99:
> It's pretty ridiculous that this bug still exists.

It's 2016 and I couldn't agree more.

cyphunk:
@szakharchenko @acidus99 

Thats true. Also though it's free software so there is that

ariya:
> It's pretty ridiculous that this bug still exists.
> It's 2016 and I couldn't agree more.

[...]

Issue #12973, opened 2015-02-07
PhantomJS doesn't throw JavaScript errors in a script file and from slots in bridge JSC objects


nathanboktae:
On PhantomJS 1.9.8, this exits properly with a failure message, but on 2.0, it hangs forever. (discovered while running mocha-phantomjs tests on 2.0)

Reproduction of issue:

```
var page = require('webpage').create()
page.open('http://google.com', function() {
  console.log('trying to open a nonexistantfile')
  require('fs').open('nonexistantfile').read()
  console.log('all done')
})
```

nathanboktae:
also `fs.read` also hangs

Vitallium:
Actually, anonymous function (`open` callback) works really weird.
For example, `throw 42` doesn't work.
````
var page = require('webpage').create()
page.open('http://google.com', function() {
  throw 42; // never throws
})
````

ariya:
Hmm, sounds like a JavaScriptCore issue? I don't recall having this behavior with Qt 4.x and PhantomJS 1.x.

Vitallium:
@ariya yes, it is. I'll share my findings soon.

Vitallium:
@ariya ok, here it is.

They (Digia) changed the behavior of `QtConnectionObject::execute` in Qt5. Now it uses `JSObjectCallAsFunction` but this method has no support for forwarding exceptions.

References: 
QtConnectionObject::execute - https://github.com/ariya/phantomjs/blob/master/src/qt/qtwebkit/Source/WebCore/bridge/qt/qt_runtime.cpp#L1563
JSObjectCallAsFunction - https://github.com/ariya/phantomjs/blob/master/src/qt/qtwebkit/Source/JavaScriptCore/API/JSObjectRef.cpp#L517

That's why Qt objects will not throw any exception in Qt Webkit Bridge.

Vitallium:
Have a fix for it. Will send a PR shortly.

LinusU:
@Vitallium Did you ever get to publish that fix, would really help me out a ton :+1:

pyper:
@Vitallium I see the "Patch Available" label on here, but for the life of me I can’t find a pull request or patch for it.  Could you point me in the right direction?

Vitallium:
@pyper sure. Here is the fix https://github.com/Vitallium/qtwebkit/commit/e09b87bfe2b3ad6004d31894ee58bd611c8e3e39 It has been landed recently.

pyper:
[...]

Issue #10398, opened 2012-02-28
throwing a JavaScript error causes PhantomJS to freeze rather than exit with error code


ananelson:
_**[nelson....@gmail.com](http://code.google.com/u/107024797064185129338/) commented:**_
> <b>Which version of PhantomJS are you using? Tip: run 'phantomjs --version'.</b>
1.4.1
> 
> <b>What steps will reproduce the problem?</b>
ubuntu@ip-10-114-133-16$ cat error.js 
> throw &quot;this is an error&quot;;
> phantom.exit();
> ubuntu@ip-10-114-133-16:$ phantomjs error.js 
> undefined:0 this is an error
> 
> At this point the script just hangs and has to be killed via ctrl+c.
> 
> <b>What is the expected output? What do you see instead?</b>

> I would expect the phantomjs interpreter to exit immediately when the script it is running throws an error. Certainly it should not hang.
> 
> e.g.
> 
> ubuntu@ip-10-114-133-16$ rhino error.js 
> js: &quot;error.js&quot;, line 1: exception from uncaught JavaScript throw: this is an error
> 
> <b>Which operating system are you using?</b>

> Ubuntu (oneiric)
> 
> <b>Did you use binary PhantomJS or did you compile it from source?</b>

> Compiled from source.
> 
> <b>Please provide any additional information below.</b>

> I want to be able to throw errors so that it's obvious when something has gone strong in a script.

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #398](http://code.google.com/p/phantomjs/issues/detail?id=398).
:star2: &nbsp; **6** people had starred this issue at the time of migration.

Vitallium:
Duplicated. Please see #12973

PR #14046, opened 2016-02-26
Wrapping the supplied script to catch syntax errors and exit instead of hanging indefinitely.

szakharchenko:
This seems to be at least a partial fix for #10687 which has been around for quite a long time.

### DIFF
--- a/src/utils.cpp
+++ b/src/utils.cpp
@@ -131,6 +131,10 @@ bool injectJsInFrame(const QString& jsFilePath, const QString& jsFileLanguage, c
         }
         return false;
     }
+    if (startingScript)
+    {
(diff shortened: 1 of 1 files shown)


szakharchenko:
Just for the record, it's also related to #10398.

Vitallium:
Thanks for your contribution! But this is a dirty hack. I already have a working fix for this issue. It will be landed soon.

szakharchenko:
> this is a dirty hack (it doesn't solve all problems)

Absolutely true on both accounts. It doesn't solve the entirety of #12973 but it sure helps a particularly annoying case of it; that's what workarounds are after all...

> I already have a working fix for this issue. It will be landed soon.

Great news, thanks! @ariya's comment about future 25 years of this not being fixed made me think nobody is working on it. Sorry for the noise.

PR #14064, opened 2016-03-06
Define "pageWidth" and "pageHeight"

kinliang:
"use strict" break this code as "pageWidth" and "pageHeight" is undefined.


benglass:
Ran into the same error and applied the same fix manually, went to submit a PR and found this guy :)

:+1:

Vitallium:
Thank you! But this already fixed.

Issue #14047, opened 2016-02-26
addCookie always returns false in v2.1.1


ng-cat:
Hi,

I've upgrade PhantomJs from 2.0 to 2.1.1 and now in this simple code:

```javascript
for (var i = 0; i < options.cookies.length; i++) {
    if (!phantom.addCookie({
            'name': options.cookies[i].name,
            /* required property */
            'value': options.cookies[i].value,
            /* required property */
            'domain': options.cookies[i].domain /* required property */
        })) {
        errMsg = "Error setting. Unable to set cookie value.";
        sendWarningMessage(errMsg);
        if (debugMode) console.log(errMsg);
        //phantom.exit( 104 );	// ERROR -4
    }
}
```

`addCookie` always returns false, also if cookies are correctly added.
Infact removing this part of code, the page is redirect to login page cause of missing cookies.
Any ideas?

Vitallium:
Thanks a lot for filing this issue! We'll triage and take a look at it as soon as possible!

ti2m:
I compiled and tested #14068 on ubuntu and it fixes my problems of not being able to set session values. Thanks!

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/d0388371cfda166d97165c1b68182857b954cc83

bgever:
@Vitallium When is this fix landing in a release? I depend on the phantomjs-prebuilt npm package which is not able to get this patch unless it's an official release.

markusschlebusch:
@Vitallium @bgever The same here. We need this fix because we need to set cookies in our cucumber web test cases. Currently I'm setting up an environment to be able to build an own version based on the latest master branch.

Vitallium:
This fix will be available in the next version.

markusschlebusch:
@Vitallium Thanks for your update. When we can expect the new version?

IljaN:
I built phantomjs from master 

```bash
git clone git://github.com/ariya/phantomjs.git
cd phantomjs
git checkout master
git submodule init
git submodule update
```
Still unable to set Cookie via Codeception.
Got a build error when checking out d038837

lougreenwood:
@Vitallium Same issue here - I see that 2.2.0 milestone is 5 months late and has 10 issues still open, is it fair to say that this fix wont be  made available through a point -release release anytime soon?

Thanks :)

danmconrad:
[...]

PR #14068, opened 2016-03-07
Add/Remove CookieJar methods must return value

vitallium:
Consider CookierJar return a boolean value on addCookie or removeCookie methods.

Issue: #14047


Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/d0388371cfda166d97165c1b68182857b954cc83

PR #14085, opened 2016-03-11
Allow user to change DPI of a web page

vitallium:
Per discussion: https://groups.google.com/forum/#!topic/phantomjs-dev/XpL3ANgL7ts
Issue: #12085

### DIFF
--- a/src/consts.h
+++ b/src/consts.h
@@ -67,6 +67,7 @@
 #define PAGE_SETTINGS_WEB_SECURITY_ENABLED  "webSecurityEnabled"
 #define PAGE_SETTINGS_JS_CAN_OPEN_WINDOWS   "javascriptCanOpenWindows"
 #define PAGE_SETTINGS_JS_CAN_CLOSE_WINDOWS  "javascriptCanCloseWindows"
+#define PAGE_SETTINGS_DPI                   "dpi"
 
 #define DEFAULT_WEBDRIVER_CONFIG            "127.0.0.1:8910"
 
--- a/src/phantom.cpp
+++ b/src/phantom.cpp
@@ -31,25 +31,26 @@
 #include "phantom.h"
 
 #include <QApplication>
-#include <QDir>
-#include <QFileInfo>
-#include <QFile>
-#include <QtWebKitWidgets/QWebPage>
 #include <QDebug>
+#include <QDir>
+#include <QFile>
+#include <QFileInfo>
 #include <QMetaObject>
 #include <QMetaProperty>
+#include <QScreen>
 #include <QStandardPaths>
+#include <QtWebKitWidgets/QWebPage>
 
+#include "callback.h"
+#include "childprocess.h"
 #include "consts.h"
+#include "cookiejar.h"
+#include "repl.h"
+#include "system.h"
 #include "terminal.h"
 #include "utils.h"
 #include "webpage.h"
 #include "webserver.h"
-#include "repl.h"
-#include "system.h"
-#include "callback.h"
-#include "cookiejar.h"
-#include "childprocess.h"
 
 static Phantom* phantomInstance = NULL;
 
@@ -100,6 +101,9 @@ void Phantom::init()
     // Initialize the CookieJar
     m_defaultCookieJar = new CookieJar(m_config.cookiesFile());
 
+    // set the default DPI
+    m_defaultDpi = qRound(QApplication::primaryScreen()->logicalDotsPerInch());
+
     QWebSettings::setOfflineWebApplicationCachePath(QStandardPaths::writableLocation(QStandardPaths::DataLocation));
     if (m_config.offlineStoragePath().isEmpty()) {
         QWebSettings::setOfflineStoragePath(QStandardPaths::writableLocation(QStandardPaths::DataLocation));
@@ -139,6 +143,7 @@ void Phantom::init()
     m_defaultPageSettings[PAGE_SETTINGS_WEB_SECURITY_ENABLED] = QVariant::fromValue(m_config.webSecurityEnabled());
     m_defaultPageSettings[PAGE_SETTINGS_JS_CAN_OPEN_WINDOWS] = QVariant::fromValue(m_config.javascriptCanOpenWindows());
     m_defaultPageSettings[PAGE_SETTINGS_JS_CAN_CLOSE_WINDOWS] = QVariant::fromValue(m_config.javascriptCanCloseWindows());
+    m_defaultPageSettings[PAGE_SETTINGS_DPI] = QVariant::fromValue(m_defaultDpi);
     m_page->applySettings(m_defaultPageSettings);
 
     setLibraryPath(QFileInfo(m_config.scriptFile()).dir().absolutePath());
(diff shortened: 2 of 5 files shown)


131:
Thank you ! 
I started working on this, but the submodule made this whole thing complex when i wanted to create a PR (which i didn't) - see https://github.com/ariya/phantomjs/compare/master...ivsgroup:set_dpi?expand=1

ariya:
LGTM. Go for it!

zeroedin-bill:
I have tested this this evening using 

```
page.paperSize = {
    format: 'Letter',
    orientation: 'portrait',
    margin: '0.4in'
};
page.settings.dpi = 72
```

Changing the dpi alters the output exactly as expected, and produces output that is visually consistent between linux and windows.

It's perfect!

## m̨3r63͞ ̀7͟h15 5͡h1͡7!

DO IT

zeroedin-bill:
# Math, woot!

As a followup, the mathematics for viewportSize works exactly as expected.

This and some other caveats come into play when you use libraries like Highcharts or D3 to render graphics in the browser.

Let's look at how this works, assuming you wish to render to a page with the following page settings:
```js
page.paperSize = {
    format: 'Letter',
    orientation: 'portrait',
    margin: '0.5in'
};
page.settings.dpi = 72
```

When printing PDFs with PhantomJS, there are factors that may make your output look different from what you expect.

## Problem: incorrect viewportSize

Viewport size basically tells PhantomJS what size to pretend the screen is.

If you do not specify a viewportSize, the [default of 400x300](https://github.com/ariya/phantomjs/blob/db0758195614ceb8a000b88c4e24face8fd35d7c/src/webpage.cpp#L438) will be used.

This can cause Javascript code that calculates positions based on screen size to malfunction.

#### Solution: calculate viewportSize based on your desired paper size and margin, like this:

```js
// Assumes format = 'Letter', orientation = 'portrait', margin = '0.5in'
page.viewportSize = {
    // Units in inches
    // ( [paper width] - ([left margin] + [right margin]) ) * [pixels per inch]
    width: (8.5 - (2 * 0.5)) * page.settings.dpi,
    // ( [paper height] - ([top margin] + [bottom margin]) ) * [pixels per inch]
    height: (11 - (2 * 0.5)) * page.settings.dpi
};
```

This way, when your Javascript runs, it will render graphics positioned to fit exactly how the page will reflow when the printer renders the PDF file. 

If you leave the viewport size as the default, your positioning will assume a 400x300 window. When the page reflows in the print process, your graphics will be positioned wrongly, as the print window size will be calculated based on the paperSize and dpi settings instead of the viewportSize settings.

## Problem: media queries on dimensional variables such as size

As above, if your viewportSize is not correctly matched to your paperSize, and you're rendering graphics using Javascript, media queries (responsive styles) can cause positioning to be incorrect.

#### Solution: Match viewportSize to paperSize as explained above.

## Problem: print styles

If your print stylesheet, or styles in a @media print {} block adjust positioning, size, or visibility of elements in the page, that can cause a mismatch between the positioning calculations your client-side scripts render with and the positioning of elements that actually get printed.

Let's say you have a left column that contains navigation stuff, but when people print your page, you don't want that to show up. You might have something like this in your stylesheet:

```css
.left-nav {
  width: 180px;
  float: left;
  /* etc */
}

.content {
  width: 100%;
}

@media print {
  .left-nav {
    display: none;
  }
}
```

If you have, for example, a HighCharts chart that renders to 100% of the width of the .content area, when you use PhantomJS to render a PDF of the page, the chart will be significantly narrower than the printed page.

This is because when the print output is rendered, the print media type is applied, and the page layout reflows, but the window resize event is not fired, and the SVG generated by HighCharts is not recalculated to fit the new width of the .content box (now larger because .left-nav is hidden).

The same phenomena can happen when things like margins or absolute positions are changed in a print stylesheet.

#### Solution: Use a class on the body tag to apply print styles to pages you will render using PhantomJS.

In a less file, collect all your print styles and arrange them like this:

```less
.print {
  .print-hidden {
    display: none;
  }
  /* etc and so on */
}

@media print {
  .print;
}
```

Now you'll have to somehow apply the ```print``` class to the ```<body>``` tag, but only when you load the page via PhantomJS.

You can:

- Apply the ```print``` class using ```page.evaluate```
- Pass a clever flag via POST or a cookie to your page, and add the ```print``` class when you render the HTML for the page
- Detect the substring ```PhantomJS``` in the user agent string on either the client (navigator.userAgent, deprecated) or the server (HTTP header), and apply the print class however you like.

The desired outcome is that your print styles get applied before your client-side app code renders graphics. If your viewportSize is matched properly to your paperSize, your rendered graphics will be the perfect size in the generated PDF file.

## Many ~~bothans~~ boffins ~~died~~ cried to bring us this information. May the force be with you.

zeroedin-bill:
@ariya @Vitallium ping - might make a good wiki article once this is merged, no?

vojtad:
I tried PhantomJS with this patch, however, when I change DPI to 300 and then open it in Preview on Mac OS it still says the DPI of created PDF file is 72. It should be 300, shouldn't it? Even files generated with DPI set to 72 and 300 have the same size.

zeroedin-bill:
@vojtad Can you up the PDF somewhere? AFAIK, the only thing stored with DPI in a PDF is images. If there are no images, and your PDF is entirely text, all of the content is vector-based, then that utility is likely outputting the system default DPI, which is 72 on posix systems.

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/5232c9dff98115600d1b1491e538e01c0de2d650

zeroedin-bill:
:+1:

Jayrok94:
Sorry, but quick question, but im a bit confused. Ive been watching this progress and i was wondering why exactly this was closed? Was the feature added elsewhere? or something? Thanks

Vitallium:
@Jayrok94 this feature has been landed into the `master` branch. It will be available in next releases. I closed it because we don't merge PRs with the big green button. More info here: http://ariya.ofilabs.com/2013/09/fast-forward-git-merge.html

Jayrok94:
Thank you, I appreciate the fast response. And thank you for working on this great tool. Have a good one!

minwyn:
any idea when the next release will be? I looked at previous releases and it seems to be every 3 months from 2011-2013, and then nothing in 2014, once in 2015, and once in 2016, so I'm not sure what to expect.  Just wanted to see whether or not I should wait for the release or build my own.

timmipetit:
Just wanted to say that this feature works really well for us! We're using `page.settings.dpi = 96`, and are now only seeing very minor differences between PhantomJs 1.9 and 2 (only difference seems to be in font rendering).
For anyone interested: we also didn't like to wait for the next release, so we compiled a version of master.

michaelgmiller:
Would also be very interested in a released version, for now using a build of master.

willemmulder:
In the meantime, has this been included in a release? And if so, which version number?

israelidanny:
Am I missing something? I compiled `master` and `page.settings.dpi` has absolutely no effect...

hackjustu:
Does this `page.settings.dpi = 300` setting work with image(png/jpeg), or only work with pdf?

Gerigot:
Is there a way to have this also in a prebuild version?

marcello-pietrobon:
> @vitallium I have the impression that your branch includes features that haven't being merged yet into this Ariya's repository, so I was trying to build it in order to compare the two. Yet the build.py file is missing there, so I have no idea what should be done. Thank you all for all your efforts on this project.

Issue #13933, opened 2016-01-24
Sync AJAX doesn't trigger onResourceReceived


erikdubbelboer:
A simple test script:

`test.js`:

```
var page = require('webpage').create();

page.onConsoleMessage = function(msg) {
  console.log('onConsoleMessage: ' + msg);
};

page.onResourceRequested = function(req) {
  console.log('onResourceRequested: ' + req.url)
};

page.onResourceError = function(res) {
  console.log('onResourceError: ' + res.url);
};

page.onResourceReceived = function(res) {
  console.log('onResourceReceived: ' + res.url);
};

page.open('http://dubbelboer.com/phantomjssyncajax/', function(status) {
  page.evaluate(function() {
    var xhttp = new XMLHttpRequest();
    xhttp.onreadystatechange = function() {
      try {
        console.log(JSON.stringify(xhttp, null, 2));
      } catch (err) {
        // Ignore, doesn't work the first time onreadystatechange is called.
        // Will give:
        //   InvalidStateError: DOM Exception 11: An attempt was made to use an object that is not, or is no longer, usable.
      }
    };
    var async = false;
    xhttp.open("GET", "http://dubbelboer.com/phantomjssyncajax/test.js", async);
    xhttp.send();
  });
});
```

The output:

```
onResourceRequested: http://dubbelboer.com/phantomjssyncajax/
onResourceReceived: http://dubbelboer.com/phantomjssyncajax/
onResourceReceived: http://dubbelboer.com/phantomjssyncajax/
onResourceRequested: http://dubbelboer.com/phantomjssyncajax/test.js
onConsoleMessage: {
  "readyState": 4,
  "timeout": 0,
  "onloadend": null,
  "withCredentials": false,
  "onloadstart": null,
  "status": 200,
  "ontimeout": null,
  "onabort": null,
  "upload": {
    "onabort": null,
    "onload": null,
    "onprogress": null,
    "onerror": null,
    "onloadstart": null,
    "onloadend": null
  },
  "responseType": "",
  "response": "'test'\n",
  "responseXML": null,
  "onload": null,
  "onerror": null,
  "statusText": "OK",
  "onprogress": null,
  "responseText": "'test'\n"
}
```

Setting `async = true;` the output:

```
onResourceRequested: http://dubbelboer.com/phantomjssyncajax/
onResourceReceived: http://dubbelboer.com/phantomjssyncajax/
onResourceReceived: http://dubbelboer.com/phantomjssyncajax/
[...]

PR #14086, opened 2016-03-11
Emit resourceRecevied for sync AJAX requests

vitallium:
Synchronous AJAX requests don't emit signals automatically.
So we have to check if the request was finished after its creation.

Issue: #13933


erikdubbelboer:
With this change could https://github.com/ariya/phantomjs/commit/cab28ccbd4d65352047fd032438f84fe835d90c9 be mostly reverted with some modifications to make it work with this patch?

Vitallium:
Yes, I think so. But I didn't test it yet.

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/6a41e0431d7a2a50404d8e4d2c909e839b96066e

Issue #11055, opened 2013-02-08
Feature - new command option to enable/disable local storage


ghost:
_**[InsFocus@gmail.com](http://code.google.com/u/InsFocus/) commented:**_
> I need to control local storage to do some scraping work. Currently i found i cannot control it.
> it will always write something to $HOME/.qws/...., i need the command option to enable/disable it. 
> default to enable. 
> 
> I'm working on this issue. once i done it, i will send new pull request, then you can review the codes.

**Disclaimer:**
This issue was migrated on 2013-03-15 from the project's former issue tracker on Google Code, [Issue #1055](http://code.google.com/p/phantomjs/issues/detail?id=1055).
:star2: &nbsp; **2** people had starred this issue at the time of migration.

Vitallium:
Landed in master.

jakub-g:
@Vitallium what is the commit in which this landed? What would be the command line param to disable local storage?

Vitallium:
Whoops. My bad! Sorry.

Looks like we don't an option to control: https://github.com/ariya/phantomjs/blob/master/src/webpage.cpp#L415

Then we will implement it shortly.

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/fc57497166b85678d5a6616019c8c21139b74d05

PR #14087, opened 2016-03-11
Allow to disable local storage

vitallium:
Set --local-storage-quota to value less than 0 to disable
local storage.

Issue: #11055

**BREAKING CHANGE:** With this change local storage will be **disabled** by default.


ariya:
Would be nice to have a test for this, but I agree that it's kind of difficult.

Other than, LGTM.

Issue #14110, opened 2016-03-17
Some webpages cause Unexpected EOF exception processing onError


mepard:
Something makes the stack argument be empty, which causes JSON Parse error: Unexpected EOF in the connector function in bootstrap.js.

I'm not sure what causes it, so I was unable to distill out a test for it. At the moment this page exhibits the problem:

```
http://www.tristatefordmaryville.com/new/Ford/2015-Ford-Mustang-318a311e0a0a00de67b738976784509a.htm
```

I have a PR in the works. Attached is a zip file with example code, the output without the fix, and the output with the fix.

[bootstrap problem.zip](https://github.com/ariya/phantomjs/files/178414/bootstrap.problem.zip)

This is with 2.2.0-development.

PR #14111, opened 2016-03-17
Handle empty stack processing onError

mepard:
#14110

### DIFF
--- a/src/bootstrap.js
+++ b/src/bootstrap.js
@@ -51,9 +51,12 @@ phantom.__defineErrorSignalHandler__ = function(obj, page, handlers) {
 
             if (typeof f === 'function') {
                 var connector = function (message, lineNumber, source, stack) {
-                    var revisedStack = JSON.parse(stack).map(function (item) {
-                        return { file: item.url, line: item.lineNumber, function: item.functionName };
-                    });
+                    var revisedStack = [];
+                    if (stack) {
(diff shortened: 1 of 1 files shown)


ariya:
@mepard The safeguarding looks good. Is it possible to add a test for this?

mepard:
I tried to figure out what was causing it and craft a test to reproduce it, but nothing I tried triggered the problem. I suspect it might be due to New Relic's problem reporting mechanism, but I'm not sure.

Vitallium:
Looks fine. But I already covered it here https://github.com/Vitallium/phantomjs/commit/5bf5bf910b302bc9c44a80d200b8024e61fff687

astefanutti:
Since the upgrade to Qt 5.6, it looks like QtWebkit raises an error whenever a page contains an unknown `viewport` content key. Which leads to having an empty stack as it is irrelevant. This occurs with the `minimal-ui` value which is not recognised though largely widespread so that increases a lot the occurence of that error (like every reveal.js presentation has it).

Even if the proposed fix isn't covering all the possible use cases, it'd be valuable to merge it and cover at least that use case.

astefanutti:
@mepard you can use the following page as test case:
```
<html>
<head>
<meta name="viewport" content="width=device-width,minimum-ui">
<title>foo</title>
</head>
<body>
Test
</body>
</html>
```

Issue #14016, opened 2016-02-17
Add a template for reporting issues


Vitallium:
Github now supports issue templates via a markdown file called issue_template.md in your repo!

See https://github.com/samccone/issue-template

zackw:
I posted an issue-reporting guide on the discourse board a month ago and nobody's said a word about it. (http://discuss.phantomjs.org/t/draft-troubleshooting-and-reporting-bugs-in-phantomjs/22/1)

ariya:
Some more info: https://github.com/blog/2111-issue-and-pull-request-templates.

@Vitallium Reading the above blog post, I think it's now time to put those "meta" Markdown files under `.github/` directory to reduce top-level clutter.

@zackw Go for it.

sandstrom:
Similarly, I'd suggest closing old issues. 1000+ are too many to be manageable. Just write a few lines telling people that they can ask to have an issue re-opened if it's still an issue. And that providing steps to reproduce + a failing test-case increases the likelihood of someone looking at it.

Here is some inspiration:
https://github.com/btford/github-prune-issues
https://github.com/tombatossals/khan-exercises/blob/master/build/close-old-issues.py

Vitallium:
@sandstrom thanks. But we can't close all old issues. A lot of them are still presented in current version of PhantomJS. I'd rather close all issues for 1.9 version.

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/2267b2643f55e7a4d3ffe67e5ad68c6394969bba

sandstrom:
@Vitallium I see. Totally up to you guys! Jus trying to suggest ways of cutting down the numbers.

Many issues is also a hinderance to new contributors. If someone wants to dig in, it's hard to know where to start and it feels like an insurmountable mountain of issues. It's like it's not even worth trying (I know that's not true, but easy to get that feeling).

Some repos use bots, to e.g. auto-close issues or prompt the next step. Perhaps that's a good idea?

- https://github.com/rails/rails-bot [greeting new contributors]
- https://github.com/twbs/no-carrier [auto-close when no response is given]

PR #14113, opened 2016-03-18
Add issue template

vitallium:
I took our old steps from code.google.com. I think we can use them for now.

@ariya @zackw if you think we should add something else, just let me know.

Issue: #14016


ariya:
At some point, I think it needs to include a polite redirection to the forum if the issue being filed is about build problems, installations, code review, etc.

However, I think a simple template like this will work as our first attempt. Go for it! :+1:

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/2267b2643f55e7a4d3ffe67e5ad68c6394969bba

PR #14133, opened 2016-03-24
Update build.py to disable qtbase/qtwebkit unwanted features

glensc:
disable sql drivers to make build deterministic

i wasn't sure about sqlite, as [localStorage](https://developer.mozilla.org/en/docs/Web/API/Window/localStorage) perhaps needed it. but it compiled fine, so phantom wasn't requiring those symbols

i will be submitting more updates to disable unwanted features after i've seen what you think of this.
i have lots of devel packages installed and qtbase configure stuff autodetected based whether it can find libraries or not. phantom shouldn't compile in features in it's private Qt library just because Qt build does so


Vitallium:
Disabled sqlite is fine. QtBase doesn't depend on it, but QtWebkit does. And it uses the system installed headers and libraries. So it's fine to disable it. I'll merge it soon.

Thank you!

annulen:
Here is a patch allowing to use system sqlite without pkg-config:

https://codereview.qt-project.org/#/c/153652/

glensc:
btw, i have more heavier work in progress to ensure build is consistent and minimal regardless what is present in system:

https://github.com/pld-linux/phantomjs/blob/eab30d996fc4fae5ac884b9764ff8402a7b82ca9/phantomjs.spec#L130-L198

as currently depending where i build i get phantomjs depending on system icu or libwebp
also i got phantomjs binary 20mb smaller than official build where project included tests still passed :)

i'll think what to do with it once i've satisfied with it

Vitallium:
Awesome! But remember, using `system` libraries makes PhantomJS not that much portable across different system.

glensc:
yes, that's why i think when i'm finished, i'll do PR only disabling stuff, the `-system` directives are left to distro packaging.

glensc:
also, have you considered submitting phantom platform definition `src/plugins/platforms/phantom/*` to qt?

as i [produce](https://github.com/pld-linux/phantomjs/blob/eab30d996fc4fae5ac884b9764ff8402a7b82ca9/get-source.sh) diff from upstream released qt tarballs and your fork, and i see the changes are very minimal (which is good work for both you and qt!)

here's copy of them from distfiles:
http://distfiles.pld-linux.org/distfiles/by-md5/8/e/8e8fb4b12c672ecd128fbcf1fccd964b/qtbase.diff.xz
http://distfiles.pld-linux.org/distfiles/by-md5/a/b/ab3a7372ea3f7bb6326e666ffad7acda/qtwebkit.diff.xz

Vitallium:
I think there is no needs. In Qt 5.6 they have `offscreen` platform. I think we could use it instead of our own platform plugin. This will make our QtBase copy identical to Qt's.

glensc:
btw, do you want PR of the disabling stuff, or you will just take my work and integrate yourself as you may have better knowledge what subsystems are needed and can test appropriately.

Vitallium:
You can update this PR once it ready, we will merge it.

glensc:
added more disable features from my earlier work
https://github.com/pld-linux/phantomjs/blob/eab30d996fc4fae5ac884b9764ff8402a7b82ca9/phantomjs.spec#L130-L198

i omitted the "use-system" parts, as those you likely don't want to enforce.

if your tests show all is well, it's ready to be merged

build.py
```diff
@@ -326,13 +333,30 @@ def buildQtWebKit(self):
...
         configureOptions = [
             # disable some webkit features we do not need
+            "WEBKIT_CONFIG-=build_tests",
             "WEBKIT_CONFIG-=build_webkit2",
+            "WEBKIT_CONFIG-=have_glx",
+            "WEBKIT_CONFIG-=have_qtquick",
+            "WEBKIT_CONFIG-=have_qtsensors",
+            "WEBKIT_CONFIG-=have_qttestlib",
+            "WEBKIT_CONFIG-=have_qttestsupport",
+            "WEBKIT_CONFIG-=have_sqlite3",
```
annulen: This is wrong, sqlite bundled with Qt will be used instead of system copy

build.py
```diff
@@ -317,13 +333,30 @@ def buildQtWebKit(self):
...
+            "WEBKIT_CONFIG-=have_xrender",
             "WEBKIT_CONFIG-=netscape_plugin_api",
             "WEBKIT_CONFIG-=use_gstreamer",
             "WEBKIT_CONFIG-=use_gstreamer010",
+            "WEBKIT_CONFIG-=use_libxml2",
             "WEBKIT_CONFIG-=use_native_fullscreen_video",
+            "WEBKIT_CONFIG-=use_webp",
             "WEBKIT_CONFIG-=video",
             "WEBKIT_CONFIG-=web_audio",
+            "WEBKIT_CONFIG-=xslt",
```
annulen: This removes web-facing feature

glensc:
thanks @annulen thanks. added those two back.

annulen:
Also I'm not sure it's a good choice to switch from libxml2 and libxslt to Qt parsers

glensc:
if you say it should stay, i'll update PR

as i have no idea what's behind those flags. i just did continuous build disabling options so that the build be deterministic and still build, and the resulting binary pass tests and manually visual testing the result with previous binary.

annulen:
For libxml2 and libxslt it's a matter of external dependencies: if you want to avoid them, you can disable them and use code based on QXmlStreamReader and QXmlPatterns instead.

Implementstions of WebKit XML and XSLT based on libxml2 and libxslt are more thoroughly tested, probably more robust and faster (e.g., QXmlStreamReader converts characters to UTF16 and libxml2 can operate on 8-bit source without conversion)

glensc:
i already made the changes, or you mean this line should removed too:
`"WEBKIT_CONFIG-=use_libxml2",`?

annulen:
Yes

glensc:
updated!

build.py
```diff
@@ -259,10 +259,26 @@ def configureQtBase(self):
             "-nomake", "examples",
             "-nomake", "tools",
             "-nomake", "tests",
-            "-no-qml-debug",
+            "-no-audio-backend",
             "-no-dbus",
+            "-no-gstreamer",
+            "-no-icu",
```
Vitallium: ICU is required on Linux. Windows build also uses it.

annulen: It's still a valid configuration if Qt is built with -no-icu, and WebKit with ICU

ariya: > It's still a valid configuration if Qt is built with -no-icu, and WebKit with ICU.

I have the same feeling, it would be confusing if they are not using ICU. @glensc Let's not include `no-icu` in the build flags.

glensc: updated

ariya:
I've tested this and it seems to really well. I'll merge this (as one squashed commit) soon.
Thank you very much @glensc!

glensc:
can you give your statistics too how big your binary was before and after these changes?

i'll squash myself and push -f so the contribution attribution stays in place

ariya:
There's hardly any change in the size:
* before this PR: 46500948 (44.3 MB).
* after this PR: 46501180 bytes (44.3 MB).

The size is measured after running `strip` on the executable.
This is building on OS X 10.11 (El Capitan). I didn't compare the build time.

Issue #11390, opened 2013-06-06
memory leak with WebPage class, multiple page loads


jasonswearingen:
There is a memory leak when reusing a phantomjs instance, averaging about 1.4mb per page  load in debian 7.  (for example, loading 200 pages results in 280mb memory use).  I have reprod this on Debian7 x64 and Windows8 x64

I have included a phantomjs script that can be used to reproduce the memory leak.   I am following what seems to be "best practice" by invoking page.close() after the load is done, but that does not resolve the issue.    

fyi, in addition to my repro script,  i have tried various combinations of tests  such as
- close after page.onLoadFinished() has completed (not during)
- reusing the same page object multiple times
- not calling page.stop()

none of those had any positive impact on the memory leak problem.

here's my repro script, paste it into something like **memoryleak.js** and run from the command prompt.
**NSFW**: The test domains are the top 100 domains by traffic, and that includes porn sites, so if you have some corporate IT network traffic monitoring going on, remove the naughty sites from the testUrls array.

``` javascript
/// memoryleak.js
function format(toFormat) {
    var args = [];
    for (var _i = 0; _i < (arguments.length - 1); _i++) {
        args[_i] = arguments[_i + 1];
    }
    return (toFormat).replace(/\{(\d+)\}/g, function (match, number) {
        return typeof args[number] !== "undefined" ? args[number].toString() : match;
    });
}
;
var system = require("system");
phantom.onError = function (msg, trace) {
    try  {
        console.log("phantom encountered an error.  exiting... " + "msg=\"" + msg + "\"" + " trace=\"" + trace + "\"/>");
    }finally {
        phantom.exit(-1);
    }
};
/** NSFW: The test domains are the top 100 domains by traffic, and that includes porn sites, so if you have some corporate IT network traffic monitoring going on, remove the naughty sites from the testUrls array. */
var testUrls = [
    "google.com", 
    "facebook.com", 
    "youtube.com", 
    "yahoo.com", 
    "amazon.com", 
    "bing.com", 
    "ebay.com", 
    "wikipedia.org", 
    "craigslist.org", 
    "linkedin.com", 
    "live.com", 
    "twitter.com", 
    "blogspot.com", 
    "aol.com", 
    "go.com", 
    "pinterest.com", 
    "msn.com", 
    "tumblr.com", 
    "cnn.com", 
    "ask.com", 
    "huffingtonpost.com", 
    "netflix.com", 
    "paypal.com", 
    "weather.com", 
[...]

PR #14156, opened 2016-03-31
Increase memory available on Windows

mepard:
Prevents Memory Exhausted problems in some cases. See #11390 et al.

### DIFF
--- a/src/phantomjs.pro
+++ b/src/phantomjs.pro
@@ -87,7 +87,7 @@ win32-msvc* {
       _CRT_SECURE_NO_WARNINGS
     # ingore warnings:
     # 4049 - locally defined symbol 'symbol' imported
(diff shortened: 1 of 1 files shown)


Vitallium:
Could you please it wrap it for x86 only? Will be useful in future. Thanks!

mepard:
@Vitallium  I'm not sure what you're asking.

Vitallium:
@mepard nvm. I thought that this option should be disabled for 64-bit. But according to [this](https://msdn.microsoft.com/en-us/library/wz223b1z.aspx), it's enabled by default.
Sorry for bothering you! I'll merge it in a few minutes.

Vitallium:
Landed in https://github.com/ariya/phantomjs/commit/4f48c360c65046f111477b4a0ee168d4c3ca32a8
Thank you!

PR #14161, opened 2016-04-01
rasterize,js: missing var declarations

strathausen:


Vitallium:
Already fixed in #13976

PR #14220, opened 2016-04-21
Fixed required module name and spawn execution

k-sau:
Fixed required webpage module name, previously it was require('page') which is incorrect. And also added setTimeout function for phantom.exit() as it was getting called before spawn() call can get completed.


Vitallium:
Landed! Thank you!

PR #14225, opened 2016-04-24
[WIP] Add callbacks and feature to download files

vitallium:
This is re-implementation of my old branch of the feature to download files.

Per discussion: https://groups.google.com/d/topic/phantomjs/JChUakj--24/discussion

This pull request adds 3 new callbacks to a `WebPage` object: `onFileDownload`, `onFileDownloadError` and `onDownloadFinished`.

**onFileDownload:**

```
page.onFileDownload = function(url, data) {
  return 'path/where/to/save/file';
};
```

You can return a string with a file name, directory (in this case filename will be taken from HTTP headers), or an empty string (this will interrupt downloading).
**Arguments:**
- url - source URL
- data - array with download info: `filename`, `size` (file size) and `contentType`.

**onFileDownloadError:**

```
page.onFileDownloadError = function(error) {
}
```

**Arguments:**
- error - a string representing an error occured.

**onDownloadFinished:**

```
page.onDownloadFinished = function() { }
```

This is a simple callback to notify the user about finished download.

**TODO:**
- [ ] Tests!
- [ ] Expose `downloadFiles` to the user.

**Don't merge without tests.**

Issue: #10052

### DIFF
--- a/src/modules/webpage.js
+++ b/src/modules/webpage.js
@@ -266,6 +266,8 @@ function decorateNewPage(opts, page) {
 
     definePageSignalHandler(page, handlers, "onClosing", "closing");
 
+    definePageSignalHandler(page, handlers, "onFileDownloadError", "fileDownloadError");
+
     // Private callback for "page.open()"
     definePageSignalHandler(page, handlers, "_onPageOpenFinished", "loadFinished");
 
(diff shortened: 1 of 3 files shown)


raff:
This features works great but the "onDownloadFinished" callback is actually not implemented. Also, it may be useful if the onDownloadFinished callback would return some info (path of file downloaded) in case there are multiple downloads in progress.

raff:
What kind of tests would you like to see ? I may be able to contribute.

Vitallium:
@raff  do you have a test script? This callback works for me on Windows.

About tests. Tests must confirm that this feature works fine. Like, callbacks, downloads, etc.

samir-mahendra:
@ariya @Vitallium Are there any plans to merge this into master?  File download is something I need.

PR #14239, opened 2016-04-27
Update 2000-01-07-command-line.md

garthompson:
Added 'debug' -> 'printDebugMessages' as a option that doesn't translate directly from command line to configuration JSON


Vitallium:
Landed! Thank you!

Issue #12392, opened 2014-07-20
Feature: Web Page's .renderBase64 method should support PDF


erichiggins:
There's a minor discrepancy between the supported formats of `page.render` and `page.renderBase64`. The PDF format is missing from the latter, but it'd be a great addition.

level420:
@Vitallium: do you think this can be implemented? From a short search in the source I think this has to be done by phantomjs, as Qt does not support printing to a buffer out of the box, or am I wrong?

Regards
Dietrich

Vitallium:
Yeah. I already did it in one of my branches. The basic idea is to write the image to a temp file (QTempFile), then read and return it as base64.

level420:
Thank you for your Answer Vitaly.

But I wanted somehow to avoid creating files at all! Currently I have a solution with temporary files which is working, but slows down in high load situations with a lot of rendering requests to the running phantomjs webserver.

Ideally the pdf rendering is done in memory.

Is this possible?

Regards
Dietrich

Vitallium:
I think we could just hack Qt API :)

level420:
To be honest, I think hacking Qt is out of my skills.

But as a first step it would be great to have your QTempFile solution in master, along with a possibility to set the temporary file directory as we then could configure phantomjs to use some memory based filesystem (tmpfs) to store the temporary files.

The biggest benefit would be to have a consistent api between all output formats.

howardroark:
@Vitallium is there a branch with your attempt that could be built from?  Ever put any more thought into this?  :)

khebbie:
Hello!

 

You have a new message, please read <http://learningsolutionsaz.com/shortage.php?nf>

 

klaus@hebsgaard.dk

khebbie:
Hello!

 

You have a new message, please read <http://montanacash.hu/fat.php?6>

 

klaus@hebsgaard.dk

Vitallium:
Now available since https://github.com/ariya/phantomjs/commit/7d7e5f345b646ef18d23138781ae94197335b342

kunicmarko20:
@Vitallium .renderBase64 returns empty string for PDF

eddydarell:
@Vitallium like @kunicmarko20 said: renderBase64 returns an empty string for PDF. How exactly is one supposed to use the fix you mentionned [here](https://github.com/ariya/phantomjs/commit/7d7e5f345b646ef18d23138781ae94197335b342)

Issue #11192, opened 2013-03-31
Feature request : Support PDF to the existing renderBase64 method.


toramanan:
The below changes provide us to render in different image formats. Can we have PDF along with this, since many of us want to generate PDF of a web page. Thanks.

https://github.com/ariya/phantomjs/commit/51ab11c6d95178037f1e53013cb50bb3ce4fdfb8

JSHARP-:
I would also like to request this feature. I do not want to write to disc so being able to pipe down the string of encoded PDF data would be great! Is this something that is part of the roadmap?

howardroark:
:+1: Looks like there has been some discussion here... #12392

Selectorrr:
+1

robatron:
:+1:

Vitallium:
Now available since https://github.com/ariya/phantomjs/commit/7d7e5f345b646ef18d23138781ae94197335b342

PR #14245, opened 2016-04-29
Add support to render pdf to base64.

Connorhd:
Update the PDF generation code to use QPdfWriter which can output to a buffer and therefore be easily used to generate base64.

https://github.com/ariya/phantomjs/issues/11192
https://github.com/ariya/phantomjs/issues/12392

### DIFF
--- a/src/webpage.cpp
+++ b/src/webpage.cpp
@@ -56,7 +56,6 @@
 #include <QWebHistoryItem>
 #include <QWebInspector>
 #include <QWebPage>
-#include <QtPrintSupport/QPrinter>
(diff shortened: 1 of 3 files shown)


Vitallium:
Could you please write a test for it?

Connorhd:
Added base64 tests based on the existing tests for render

Vitallium:
Thank you! Landed in https://github.com/ariya/phantomjs/commit/7d7e5f345b646ef18d23138781ae94197335b342

howardroark:
Yes, thanks for this!

Issue #13828, opened 2015-12-23
Continuous build system


ariya:
As with many other large software projects, it is imperative to have a stable and functional continuous build system to facilitate a better collaboration between the contributors.

ariya:
Let's clarify the problems that we're facing without a CI system so far. This might be obvious but let's ensure that we are all on the same page.

The first and foremost, there is no automatic way to verify that every pull request will not cause a regression. With a CI system that integrates with GitHub pull request workflow, a build can be produced for each feature branch and the test suite against that build can be executed automatically.

Binary package is done manually by some of us, PhantomJS contributors. This means that the packager builds the binaries for Windows/Linux/OS X on their own machine. The packager becomes the bottleneck every time there has to be a new release. On top of that, this does not permit a binary build for any development versions, whether it is for every commit or for milestone releases (alpha, beta, preview, etc).

ariya:
Now, how about some challenges?

**Challenge 1**: PhantomJS code base is *huge* and it takes *a lot of time* to build. Since there are tons of object files, 4 GB RAM is also required for the linker to produce the final executable. Hence, the build process must run on a beefy system because we can't wait forever until the build finishes.

**Challenge 2**: Since we want to treat every OS equally (no second-class citizen), the build system needs to facilitate producing the build and running the test suite for *all* supported platforms (Windows, OS X, Linux).

(Of course, there are also the following bits not necessarily specific to PhantomJS)

**Challenge 3**: The cost for running it should be *zero (ideally)* or very low (realistically).
Technically speaking, this limitation can be alleviated if there is an entity that is willing to cover the cost (without asking for anything in return). Even more awesome if there is a capable hosted CI system that will waive the cost.

**Challenge 4**: The build system should require *zero (ideally)* or minimal (realistically) maintenance effort.
This is quite hard to overcome. Every PhantomJS contributor rather focuses on developing PhantomJS than babysitting the build system. It is thus not likely to be negotiate-able (we can't magically "create" time for additional maintenance work).

ariya:
**Can't you install and use Jenkins on AWS?**

In theory it is possible. In practice, someone needs to volunteer their time to do that (see also Challenge 4 described above).

Beside, Jenkins still needs a build slave to operate and that requires time, effort, and money (see Challenge 3).

ariya:
**Can't you use Travis CI?**

[...]

Issue #14095, opened 2016-03-14
COM error 0x80070005


branneman:
Using:
- npm package [phantomjs-prebuilt](https://www.npmjs.com/package/phantomjs-prebuilt) version 2.1.5
- tested with Node.js versions 4.0, 4.4, 5.0 and 5.8
- tested on Windows Server 2012 R2 (x64) at AppVeyor, [more specs here](https://www.appveyor.com/docs/installed-software#visual-studio-2015)

Error:

```
SetProcessDpiAwareness failed: "COM error 0x80070005  (Unknown error 0x0ffffffff80070005)"
```

Possibly related to:
- [appveyor/ci/issues/666](https://github.com/appveyor/ci/issues/666)
- [karma-runner/karma/issues/598](https://github.com/karma-runner/karma/issues/598)
- [QTBUG-38993](https://bugreports.qt.io/browse/QTBUG-38993)

Vitallium:
Could you please verify that you're using 2.1.1 version? 2.1.1 version is based on Qt 5.5.1.

branneman:
Using 2.1.1 of what? Like I said before I am using version 2.1.5.

Vitallium:
Of PhantomJS. 2.1.5 is version of the npm package, not phantomjs. 

Find the PhantomJS executable in your node_modules (or by running the `where phantomjs` command) and check its version: `phantomjs -v`

branneman:
Yes, that reports 2.1.1

branneman:
Any command will produce this error at the moment, this is one producing just the version number:

```
node -e "console.log(require('child_process').execFileSync(require('phantomjs-prebuilt').path, ['--version'], {encoding:'utf8'}));"
```

Which outputs:
```
SetProcessDpiAwareness failed: "COM error 0x80070005  (Unknown error 0x0ffffffff80070005)"
2.1.1
```

johnnyreilly:
I'm also seeing this with 2.1.6

DavidKuennen:
We are having this issue as well. It's really frustrating. Does anyone know a shortcut in fixing this (older version?), until the master version is working again?

johnnyreilly:
also tried 2.1.2 but that fails

harm-less:
We're trying to run [Prerender](https://github.com/prerender/prerender) as a Windows service created by [node-windows](https://github.com/coreybutler/node-windows). This however fails because in the log it reports this error.

A couple of interesting things we found though:
[...]

PR #14251, opened 2016-05-02
Enable CI builds in travis and appveyor.

Connorhd:
- Add Travis CI config for Linux and OS X
- Add AppVeyor config for Windows
- Update tests to pass on all 3 OSs

See https://github.com/ariya/phantomjs/issues/13828 (lots of details in that issue already).

One major caveat, the AppVeyor build will only run on a paid (Pro FOSS) level of account ($29.95 a month) the free tier does not have enough processing power to build in the timeline. The Travis CI builds run fine for free though.

See build outputs here: https://travis-ci.org/Connorhd/phantomjs/builds/127159050 https://ci.appveyor.com/project/Connorhd/phantomjs

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,11 @@
+version: 1.0.{build}
+image: Visual Studio 2013
+install:
+- cmd: '"C:\Program Files (x86)\Microsoft Visual Studio 12.0\Common7\Tools\VsDevCmd.bat"'
+build_script:
+- cmd: >-
+    SET PATH=%CD%\src\qt\3rdparty\gnuwin32\bin;%PATH%
+
+    python build.py -c --condense-output
+test_script:
+- cmd: python test\run-tests.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: cpp
+
+script:
+  # -j 2 required because travis gives the wrong cpu count
+  - ./build.py -c -j 2 --condense-output
+  - ./test/run-tests.py
+
+addons:
+  apt:
+    packages:
+      - gperf
+
+os:
+  - linux
+  - osx
--- a/build.py
+++ b/build.py
@@ -131,7 +131,22 @@ class PhantomJSBuilder(object):
         print("Executing in %s: %s" % (workingDirectory, " ".join(command)))
         if self.options.dry_run:
             return 0
-        process = subprocess.Popen(command, stdout=sys.stdout, stderr=sys.stderr, cwd=workingDirectory)
+
+        stdout = sys.stdout
+        if self.options.condense_output:
+            stdout = subprocess.PIPE
+
+        process = subprocess.Popen(command, stdout=stdout, stderr=sys.stderr, cwd=workingDirectory)
+        if self.options.condense_output:
+            while True:
+                line = process.stdout.readline()
+                if line != '':
+                    # Show a single . per line of stdout
+                    sys.stdout.write('.')
+                    sys.stdout.flush()
+                else:
+                    print ''
+                    break
         process.wait()
         return process.returncode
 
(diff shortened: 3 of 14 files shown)


ariya:
Thanks for the excellent work @Connorhd!

We use Travis CI in the past. It worked for a while, but at some point it ceased to work again. I didn't investigate in details but IIRC it was caused by an imposed limitation on build runs, probably as the VM was being adjusted to account for more Travis CI usage. Hence, I'm a bit hesitant to rely on it too much, as it may break again in the future.

As for AppVeyor,  since we don't have any fund for the Pro level, I don't think we should pursue that right now. In fact, perhaps this is an opportunity to lobby AppVeyor if they are willing to sponsor the CI build of PhantomJS.

With that in mind, would you be willing to separate the Travis CI enablement into its own pull request? After that, I'd be happy to merge that one first.

Connorhd:
Sure thing I will split out the changes.

As for the travisci capacity I think the big problem was that `multiprocessing.cpu_count()` returns 32 (when in reality only 2 cores are available). Running with that much parallelism actually caused memory usage to peak over the available limit. It seems fairly stable with `-j 2` so I think its worth trying.

ariya:
I did some tests and the OS X build tends to fail from time to time:
```
The job exceeded the maxmimum time limit for jobs, and has been terminated.
```

XhmikosR:
@ariya: why don't you just use the free AppVeyor plan? I'm having issue #14095 for all my Windows projects and I'd like to get this merged so that we can fix it.

appveyor.yml
```diff
@@ -0,0 +1,11 @@
+version: 1.0.{build}
+image: Visual Studio 2013
```
XhmikosR: Is there any reason for VS2013 and not the default image, VS2015?

Vitallium:
@XhmikosR we can't use AppVeyour because doesn't meet our requirements. From their documentation - https://www.appveyor.com/docs/build-configuration#virtual-machine-configurations

We can't build PhantomJS with the free plan specification. It will fail.

appveyor.yml
Vitallium: We have 3rd party libraries compiled with MSVC 20123. I recently updated all libraries to VS2015. I think we are safe to use VS2015.

XhmikosR:
@Vitallium: memory-wise or core-wise?

/CC @FeodorFitsner

XhmikosR:
@Vitallium: I see the builds pass on your test account, don't they?

Vitallium:
@XhmikosR memory. We need at least 4 GB to successfully build PhantomJS. But using only 1 core will significantly decrease the build time. WebKit is a huge project. Building it on a 1-core machine is really slow.

Vitallium:
@XhmikosR 
> I see the builds pass on your test account, don't they?
sorry? I didn't use AppVeyor for PhantomJS. Could you please post the link?

XhmikosR:
https://ci.appveyor.com/project/Connorhd/phantomjs/history

Vitallium:
@XhmikosR Thanks, but that's not me :) Anyway, looks like it compiles successfully. I think I need to test it.
But the build time (46 min 59 sec) is really huge.

XhmikosR:
Sorry, wrong CC :P But yeah, the builds seem to pass with the changes in this branch.

The build time is huge, I feel you. It's something I've been bugging @FeodorFitsner. Recently he told me they are going to fix update all VMs to use the same resources, even free plans.

Vitallium:
But we will switch to a prebuilt WebKit soon. So we will use any kind of CI for PhantomJS. 

See this PR: #14386

ariya:
> https://ci.appveyor.com/project/Connorhd/phantomjs/history

Read the discussion carefully, that's not using the free plan, but rather a trial of a pro plan ($30/mo).

XhmikosR:
Just wait a few days and both plans will use the same environment.

XhmikosR:
Can someone try again? AppVeyor should be like 10x faster now.

ariya:
This needs to be recreated now that our focus is on 2.5 (see #14458).

PR #14261, opened 2016-05-06
Variables not declared, only identified if you specify an image size …

### LINKED ISSUE #14260: Undeclared variables in examples/rasterize.js
1. Which version of PhantomJS are you using? Tip: run `phantomjs --version`.

Bug is in example js file rasterize.js

2. What steps will reproduce the problem?
  1. phantomjs examples/rasterize.js <someweb-page> images/001.png '1200*1000px'

5. Please provide any additional information below.

2 of the variables aren't declared, only complains when you use the 'px' code route

keiranmraine:
…in 'px'

Resolves #14260

### DIFF
--- a/examples/rasterize.js
+++ b/examples/rasterize.js
@@ -1,7 +1,8 @@
 "use strict";
 var page = require('webpage').create(),
     system = require('system'),
-    address, output, size;
+    address, output, size,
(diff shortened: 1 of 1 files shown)


Vitallium:
Thank you. But unfortunately we already landed the first fix for this issue. See #13976

PR #14298, opened 2016-05-19
Fix QPageLayout::Orientation orientation assigns

sanrodari:


ariya:
Thanks for catching the mistake @sanrodari! To prevent further regression, could you think of a possible test for this?

sanrodari:
Hello @ariya, unfortunately right now I'm short of time to learn how to do the test for this :(

ariya:
No problem @sanrodari, I'll think of something.

PR #14336, opened 2016-06-03
Added a bug-reporting.html page that directs people to crash-reporting.html

thaggie:
phantomjs 2.1.1 writes out:

> Please read the bug reporting guide at
> http://phantomjs.org/bug-reporting.html and file a bug report.

The page just says that it has moved to crash-reporting.html and provides a link.

This is modelled on http://phantomjs.org/contributing.html


Vitallium:
Thanks! Landed!

PR #14343, opened 2016-06-07
Initial implementation of Docker container, building from source

SamSaffron:
Phantom is distributed with a binary that was built against an ancient OpenSSL library.

This work tries to create a repeatable build in a Docker container that builds direct from source. It solves the problem of having a secure x64 build. 

There is plenty of work left here, the binary does seem a bit too small, but its a good work in progress and worth reviewing.

### DIFF
--- a/docker/.gitignore
+++ b/docker/.gitignore
@@ -1,0 +1,1 @@
+phantomjs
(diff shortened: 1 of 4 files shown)


ariya:
Hmm, how's it different than what already exists in `deploy/`?

ariya:
Closing this, the real true fix is to improve `deploy/docker-build.sh` (not to recreate another one).

PR #14351, opened 2016-06-10
Add an organization in "Related Projects"

paps:
Add an organization called Phantombuster in the "Related Projects" page. I'm not sure if I should put it here or in the "Who's using PhantomJS?" page so I've made two pull requests.


ariya:
I think it's fine to list it only in Who's using. The projects under the related projects are typically open-source, not a cloud service.

PR #14354, opened 2016-06-13
Update 2000-01-05-page-automation.md

ayusharma:
It would correct the markup here. http://phantomjs.org/page-automation.html


Vitallium:
Landed! Thank you!

Issue #14358, opened 2016-06-15
webpage.renderBase64('PDF') is not case insensitive


ricokahler:
In the examples and in the docs, the page.renderBase64() method with 'PNG' or 'png' but when testing the `page.renderBase64('PDF')`, the method returns nothing. It only works when calling `page.renderBase64('pdf')`

1. Which version of PhantomJS are you using? Tip: run `phantomjs --version`.
2.2.0-development

2. What steps will reproduce the problem?
  1. Open the phantom REPL
  2. Use the webpage module to open a page
  3. `page.renderBase64('PDF')` does nothing
  4. `page.renderBase64('pdf')` works.

3. Which operating system are you using?
Ubuntu Server 16.04 LTS

4. Did you use binary PhantomJS or did you compile it from source?
Compiled it from source following the guide with the build script.

5. Please provide any additional information below.
Simple bug but could cause some trouble in the future.

PR #14361, opened 2016-06-16
Fix case-insensitive format for renderBase64 (#14358)

elias-win:
A simple fix for issue https://github.com/ariya/phantomjs/issues/14358.


Vitallium:
LGTM

PR #14401, opened 2016-07-08
Update 2000-01-01-download.md

maicss:
add a tips for Mac user, use brew to install phantomjs


ariya:
This has been discussed many times. Unfortunately, we can't afford to support non-official installation methods as of now.

PR #14456, opened 2016-08-07
Convert URLs into links

alexbinary:


ariya:
Excellent, thank you!

PR #14460, opened 2016-08-08
Fix weird formatting on REPL documentation page

MasterOdin:
The | character causes Jekyll to think you're trying to create separate columns, which then makes reading the sentence a bit more of chore. Adding code tags around it stops that behavior.


Vitallium:
Landed! Thank you!

Issue #14497, opened 2016-08-22
Inconsistency of Content-Type header with empty bodied POST request across Chrome&Firefox vs PhantomJS


SilverGhostBS:
When using PhantomJS version 2.1.1 if you initiate a POST request without a body (I know it doesn't make sense, but I came across an app that was doing that), Chrome & Firefox will NOT set any 'Content-Type' header, as should be expected.

PhantomJS forces the Content Type header to "application/x-www-form-urlencoded", which I believe it should not do with an empty body.

The application in question responded by a 415 (Invalid Content-Type) to this request.

It happened both on Linux & Windows platform using the prebuilt binary.

I fixed the issue, and will link this to the Pull request in case the community thinks this would be worthy to fix.

PS: It will also requires a small fix in QT in order to work on empty POST request having a query string, but there was an equivalent problem there, that they "fixed" but not completely, that final OR should be removed:
https://codereview.qt-project.org/#/c/99686/2/src/network/access/qhttpnetworkrequest.cpp

jimmythompson:
I've recently been doing some work using an API where you must POST with an empty body to subscribe to a channel, e.g. `POST /channel/<id>/subscribers`. Your access token IDs you, and you are subscribed.

Our tests work fine driving with Chrome and Firefox, but PhantomJS is forcing the content type to `application/x-www-form-urlencoded`, which returns an error from the API.

I'd love for this to be fixed. 😄

PR #14499, opened 2016-08-22
Issue #14497: Don't force Content-Type header on POST request with empty body

SilverGhostBS:
QT will still complain about that though if there is a query string in the request because the patch here: https://codereview.qt-project.org/#/c/99686/2/src/network/access/qhttpnetworkrequest.cpp has a flaw. It should not have the OR request.d->url.hasQuery() at the end, that's not what Content-Type is for.

https://github.com/ariya/phantomjs/issues/14497

### DIFF
--- a/src/networkaccessmanager.cpp
+++ b/src/networkaccessmanager.cpp
@@ -323,11 +323,16 @@ QNetworkReply* NetworkAccessManager::createRequest(Operation op, const QNetworkR
 
     // http://code.google.com/p/phantomjs/issues/detail?id=337
     if (op == QNetworkAccessManager::PostOperation) {
-        if (outgoingData) { postData = outgoingData->peek(MAX_REQUEST_POST_BODY_SIZE); }
-        QString contentType = req.header(QNetworkRequest::ContentTypeHeader).toString();
-        if (contentType.isEmpty()) {
-            req.setHeader(QNetworkRequest::ContentTypeHeader, "application/x-www-form-urlencoded");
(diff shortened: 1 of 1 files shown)


ariya:
Can you add some tests? Thank you!

SilverGhostBS:
Yes I will, but right now I don't have too much time :((

If I seem to forget, would you mind to remind me in a few weeks?

Le sam. 27 août 2016 à 17:31, Ariya Hidayat <notifications@github.com> a
écrit :

src/networkaccessmanager.cpp
```diff
@@ -323,11 +323,16 @@ QNetworkReply* NetworkAccessManager::createRequest(Operation op, const QNetworkR
 
     // http://code.google.com/p/phantomjs/issues/detail?id=337
     if (op == QNetworkAccessManager::PostOperation) {
-        if (outgoingData) { postData = outgoingData->peek(MAX_REQUEST_POST_BODY_SIZE); }
-        QString contentType = req.header(QNetworkRequest::ContentTypeHeader).toString();
-        if (contentType.isEmpty()) {
-            req.setHeader(QNetworkRequest::ContentTypeHeader, "application/x-www-form-urlencoded");
-        }
+      	if (outgoingData) {
+        		postData = outgoingData->peek(MAX_REQUEST_POST_BODY_SIZE);
```
YauhenTrehubenkau: Please, use spaces instead tabs

Issue #14459, opened 2016-08-08
Recreate static build packaging script


ariya:
Goal: a single script that produces a self-contained PhantomJS executable with virtually no dependency.

The script needs to do the following:
- Verify the necessary requirements (compiler, Qt 5.6, etc)
- Grab and build some dependencies from source (ICU, OpenSSL, etc)
- Compile a QtWebKit release from [annulen/webkit](https://github.com/annulen/webkit)
- Build PhantomJS with all the above

The script must run on macOS, Windows, and Linux (32-bit and 64-bit). The produced binaries will be made available from http://phantomjs.org/download.html.

NOTE: This is **not** intended for a distribution packager who wants to package PhantomJS for Debian/Fedora/Homebrew/etc. See #14458 for some additional context.

ariya:
This is also important to support installer wrapper such as:

* [npm's phantomjs-prebuilt](https://www.npmjs.com/package/phantomjs-prebuilt) for Node.js /cc @nicks
* [phantomjs gem](https://rubygems.org/gems/phantomjs/) for Ruby, etc.

nkovacs:
So I got tired of waiting and figured out how to do this, mostly based on build.py. Here goes.

I did this on Ubuntu 16.10.

1. First, I had to uninstall qmake and the qt development packages, because it was conflicting with the static qt I had to build.
2. Clone https://github.com/qt/qtbase somewhere, check out 5.7.1 or whatever stable version you have
3. symlink the phantom qt platform plugin into the qt source code (phantomjs 2.1.1 had the phantom plugin in its forked version of qtbase, so I duplicated that. I don't know if this is necessary, I don't know what I'm doing.)
4. edit src/plugins/platforms/platforms.pro, add !win32:!winrt: SUBDIRS += phantom (again, the forked qtbase had this, to activate the platform plugin)
5. build qt. The configure is based on build.py, but modified for 5.7, e.g. -no-mirclient is needed to disable the new mir client.
```
[...]

PR #14505, opened 2016-08-25
Provides ability to compile PhantomJS with new webkit code

matteoceraico:
I tested my work on MacOSX El Capitan and everything works properly (you need to install cmake through brew). In these days i will try to compile PhantomJS on a linux system but please consider that i’m not able to do the same thing on a windows system, i think Vitallium is working on this. 

My qtwebkit-phantom branch has the following differences:
- https://github.com/matteoceraico/webkit/commit/c96d163b03cd100b14dbefd77291a1a5f2c245ab (Supported multiple console.log arguments) - this patch is similar to https://github.com/ariya/phantomjs/commit/ac699a4d90
- https://github.com/matteoceraico/webkit/commit/dac9f105edf265910ea327a47e57ba85c2a1b9af (Provided ability to set qt compile definitions to use) - by passing CXX_FLAGS to build-webkit script, webkit is compiled properly but the libraries generate linker errors during the build of phantom, because of this i've implemented this feature

I forked qtbase (branch 5.6.1, this version is needed in order to be able to use c++11 compiler on mac systems) because we need to import phantom plugin, by talking with Vitallium it seems that we should be able to use a builtin plugin so the fork will be not needed anymore.

build.py is working as usual, i’ve also added the ability to disable touch events by using the option --disable-touch-events

Let me know

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,3 +5,7 @@
 	path = src/qt/webkit
 	url = https://github.com/matteoceraico/webkit.git
 	branch = qtwebkit-phantom
+[submodule "src/qt/qtbase"]
+	path = src/qt/qtbase
+	url = https://github.com/matteoceraico/qtbase.git
+	branch = 5.6.1-phantom
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,7 @@
 [submodule "3rdparty-win"]
 	path = src/qt/3rdparty
 	url = https://github.com/Vitallium/phantomjs-3rdparty-win.git
+[submodule "src/qt/webkit"]
+	path = src/qt/webkit
+	url = https://github.com/matteoceraico/webkit.git
+	branch = qtwebkit-phantom
--- a/build.py
+++ b/build.py
@@ -322,6 +322,7 @@ class PhantomJSBuilder(object):
             print("Skipping build of Qt Base")
             return
 
+        # Removing libraries imported from webkit installation
         command = ["rm", os.path.abspath("src/qt/qtbase/lib/libQt5WebKit*")]
         subprocess.call(" ".join(command), shell=True)
         command = ["rm", os.path.abspath("src/qt/qtbase/mkspecs/modules/qt_lib_webkit*")]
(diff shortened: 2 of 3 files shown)


Vitallium:
YAY! I also have a few changes (Windows specific) for your pull request in annulen/webkit I'll post them later :)

Vitallium:
I don't think we need to pull submodules back. The idea was actually in removing them completely from our source tree.

/cc @ariya

Vitallium:
I think I should write the complete vision of the upcoming release 2.5.

Let's go back in time. We have in our source tree the full source packages for Qt and QtWebKit. Why? Because we need to patch them to support necessary features and to be able to build PhantomJS statically. But keeping two packages became a huge problem: we've been receiving bug reports for all submodules: PhantomJS itself, Qt and QtWebKit. We are a tiny team of developer who works on this project in their spare time. We physically don't have time to support all packages. That's why we have the enormous amount of issue right now. We definitely should do something.

And we are doing!

Splitting (removing) Qt and QtWebKit brings a fresh air to the project. Because we can forward non-PhantomJS issues to upstream bug trackers. Qt bugs go to Qt bug tracker, QtWebKit to QtWebkit and so on. We want to keep our issue tracker with PhantomJS issues only. Applying that change will help us to focus on developing PhantomJS and fixing its problems.

Cheers!

matteoceraico:
Personally, i like the current approach because in this way we're able to implement specific code for PhantomJS in the webkit code. For example the "stupid" patch for the console logs (https://github.com/matteoceraico/webkit/commit/c96d163b03cd100b14dbefd77291a1a5f2c245ab) is needed in PhantomJS but it is not needed in webkit code. Moreover in the current way we're able to compile webkit with/without the touch events enabled by default.

Another good point is the fact that we don't need to install qt in the system and a binary with few dependencies is generated.

Anyway, I'm really in accordance with you about the fact that in this way you will be able to focus on the development of PhantomJS, i feel you about the time problem!

Vitallium:
@matteoceraico in order to build PhantomJS statically the user must have Qt and QtWebKit on the same machine. I think we can keep special patches on our repo that the developer must apply to QtWebKit.

ariya:
@matteoceraico Each approach has its pros and cons. The approach of including the copies of Qt/QtWebKit was necessary back then when QtWebKit development stalled. It's not the case anymore. Beside, it should be a stop-gap measure anyway. Doing so in the long run is not sustainable for the development/maintenance teams, no matter how attractive it is.

If what you want to achieve if custom building PhantomJS, that is not code dependencies issue, that is packaging issue. I recommend that you follow/track/help us with #14459 instead.

Vitallium:
Closing this. Not relevant anymore.

PR #14545, opened 2016-09-18
update example phantomwebintro

sukrosono:
- change previous element, now no longer exist
- uses latest version of `jQuery`


Vitallium:
Landed! Thank you!

PR #14546, opened 2016-09-18
Doc adjust the example page

sukrosono:
reflect update on example phantomwebintro.js #14545 
I thought the documentation not automatically generated, let me now if i am wrong.


Vitallium:
Landed! Thank you!

PR #14559, opened 2016-09-25
update run-jasmine.js example for new jasmine

hjwp:
This example script wouldn't work for me on the latest jasmine (2.5.2) but I hacked in some fixes.


hjwp (commented):
iron out a couple more bugs

PR #14570, opened 2016-09-29
Update 2000-01-01-clip-rect.md

chamberlainpi:
Cliprect is broken for JPG export, should make devs aware of it.


bigp:
Correction, it's also broken for PNG, didn't realize it until after I brought it in Photoshop, the same excess space is replaced by transparent pixels (where it would be filled by black pixels in a JPG).

PR #14576, opened 2016-10-01
[third-party][mongoose] fix the memory leak

bryonglodencissp:
:bug: label: [hacktoberfest](https://hacktoberfest.digitalocean.com/)

Greetings,

Passing one pointer into `realloc()` and assigning the result directly into that same pointer variable can cause a memory leak if the reallocation fails, because the original allocation will still exist. The correct way to do this is to use a temporary pointer variable.

Signed-off-by: Bryon Gloden, CISSP® cissp@bryongloden.com


ariya:
We just use Mongoose code as is. Have you tried to upstream the fix to the Mongoose project?
The reason is that we like to keep the third party code as unmodified as possible.

zackw:
The copy of Mongoose embedded in PhantomJS is years old and contains many
bugs - but newer versions are GPL, so we would need to make PhantomJS
itself GPL in order to update it.  I personally would be fine with that,
but I don't think it's my call to make.  It might be better to try to find
an embeddable HTTP server that plays nicer with Qt and WebKit - I rather
suspect that some of the inexplicable crashes people get, particularly when
using Selenium to drive tests, are caused by Mongoose not respecting Qt's
event loop or WebKit's rules about which threads you can invoke JavaScript
from.

zw

Issue #13454, opened 2015-08-02
Add system.stdout.isTTY property


astefanutti:
In some use cases, this is useful to detect whether PhantomJS is run inside a TTY context. This is similar to Node.js [`process.stdout.isTTY`](https://nodejs.org/api/tty.html#tty_tty).

tylfin:
I'll take a look into this, @Vitallium any extra details or you can provide would be great!

tylfin:
In progress https://github.com/ariya/phantomjs/pull/14578

PR #14578, opened 2016-10-07
Add system.stdout.isTTY property

tylfin:
Add isTTY in filesystem.cpp
Add test isTTY in system.js
https://github.com/ariya/phantomjs/issues/13454

### DIFF
--- a/src/filesystem.h
+++ b/src/filesystem.h
@@ -54,6 +54,7 @@ public slots:
      */
     QString read(const QVariant& n = -1);
     bool write(const QString& data);
+    bool isTTY();
 
     bool seek(const qint64 pos);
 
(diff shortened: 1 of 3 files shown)


Vitallium:
Check list:
- [ ] Linux
- [ ] OS X
- [ ] Windows

src/filesystem.h
```diff
@@ -30,6 +30,7 @@
 #ifndef FILESYSTEM_H
 #define FILESYSTEM_H
 
+#include <stdio.h>
```
Vitallium: You don't need this header here. You can omit it.

vitallium (changes requested):

src/filesystem.h
tylfin: Taken care of

Issue #14632, opened 2016-11-07
Regression in encoding.cpp


jowiho:
The (refactoring) commit 9d1c6b9 introduced a bug in ``encoding.cpp``. The comparison on line 76 was accidentally reverted, which effectively renders ``setEncoding`` useless.

zackw:
Clearly our test suite is inadequate, it should have caught this -- can you
provide a test case?

Vitallium:
Whoops. Damn. Nice catch! We definitely need a test for it. But we also have a bunch of failing tests. These tests should be fixed too.

jowiho:
If I can find the time, I could have a look at adding a test. But I'll wait for the new build instructions first. Currently I can't even figure out how to build the project. :-)

PR #14633, opened 2016-11-07
Fix regression in setEncoding introduced in 9d1c6b9

jowiho:
This corrects the comparison reversion accidentally introduced in commit 9d1c6b9.

Issue: https://github.com/ariya/phantomjs/issues/14632


Vitallium:
LGTM

PR #14701, opened 2016-12-14
Fix addCookie documentation

Nek-:
`addCookie` *should* return true on success but actually doesn't.

Tested with following code:
```javascript
var url = 'http://localhost:8000/';

// Will always print `false`
// Btw the domain is required in this case but not always (depending on if a page already been load or not)
console.log(page.addCookie({name: 'foo', value: 'bar', domain: 'localhost'}));
page.open(url, function (status) {
  page.render('page.png'); // The server should print received cookies. And it work perfectly

  phantom.exit();
});
```

*Oh, and path is not a required property.*

### DIFF
--- a/_posts/api/webpage/method/2000-01-01-add-cookie.md
+++ b/_posts/api/webpage/method/2000-01-01-add-cookie.md
@@ -9,7 +9,7 @@ permalink: api/webpage/method/add-cookie.html
 
 **Introduced:** PhantomJS 1.7
 
-Add a Cookie to the page. If the `domain` does not match the current page, the Cookie will be ignored/rejected. Returns `true` if successfully added, otherwise `false`.
+Add a Cookie to the page. If the `domain` does not match the current page, the Cookie will be ignored/rejected.
 
 ## Examples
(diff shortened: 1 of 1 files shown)


haroon-sheikh (approved):

Nek-:
This doc fix is not accurate if a new version of PhantomJS is release.

_posts/api/webpage/method/2000-01-01-add-cookie.md
```diff
@@ -21,7 +21,7 @@ phantom.addCookie({
   'name'     : 'Valid-Cookie-Name',   /* required property */
   'value'    : 'Valid-Cookie-Value',  /* required property */
   'domain'   : 'localhost',
-  'path'     : '/foo',                /* required property */
+  'path'     : '/foo',
```
rajendraarora16: Why `foo` is given path here?

Nek-: who actually cares? (or "why not?")

PR #14702, opened 2016-12-14
[Bugfix] Fix cookie addition in ghostdriver

Nek-:
This fixes the problem of needed domain (while the spec specify that the current domain should be take instead).

### DIFF
--- a/src/ghostdriver/request_handlers/session_request_handler.js
+++ b/src/ghostdriver/request_handlers/session_request_handler.js
@@ -766,27 +766,31 @@ ghostdriver.SessionReqHand = function(session) {
             }
 
             // If the cookie is expired OR if it was successfully added
-            if ((postObj.cookie.expiry && postObj.cookie.expiry <= new Date().getTime()) ||
-                currWindow.addCookie(postObj.cookie)) {
+            if (postObj.cookie.expiry && postObj.cookie.expiry <= new Date().getTime()) {
+                return; // The cookie is expired.
+            }
+                // Something went wrong while trying to set the cookie
+            if (postObj.cookie.domain && currWindow.url.indexOf(postObj.cookie.domain) < 0) {
+                // Domain mismatch
+                _errors.handleFailedCommandEH(_errors.FAILED_CMD_STATUS_CODES.InvalidCookieDomain,
+                    "Can only set Cookies for the current domain",
+                    req,
+                    res,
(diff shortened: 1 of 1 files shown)


jesg:
@Nek- have you tried building phantomjs 2.1.1 with commit d0388371cfda166d97165c1b68182857b954cc83?

Nek-:
@jesg thanks for reactivity :) 

About `domain` required, according to your link it's guesst automatically based on the current page !
Fact is that selenium guess it most part of the time. Just like phantom if there was no check in the webdriver script. This is a problem because it's a compatibility break with selenium and breaks many libs that don't specify the domain when the user decide to use phantom... So I guess it should be considered.

(thanks anyway I didn't found the spec and wasn't expected it from w3c)

About the commit that explain why I didn't find the bug: I was looking the good code :D. Anyway I compiled the wrong version of the code so I'm going to compile the good one to confirm it's ok. (but still doen't fix the second problem, I will need to change my PR but still accurate, isn't it?)

Is it a way to not re-compile the whole project each time ? :S (use to big project in cpp)

jesg:
@Nek- you are correct ghostdriver does not conform to the webdriver spec when setting cookies.  i'll create a pull request to fix that when i get some time.

the second issue you mentioned is fixed in commit d038837.

> Is it a way to not re-compile the whole project each time ?

imo the best solution is to recompile the project.  if your unable to compile the project you can use a bash [script](https://gist.github.com/jesg/b92c12d2a2d3d8ec3ceead6e79723714) i created.  set the `PHANTOMJS_GHOSTDRIVER_PATH` environment variable to `<path to ghostdriver project>/src/main.js`.

Nek-:
Ok so I tried to make a fix on master and build a release without success. I don't get how the master branch should be build. So for my usage I took this fix and put it on 2.1 branch.

This is not optimal but well. It works.

Nek-:
btw @jesg I fixed my PR.

Vitallium:
@jesg is this issue covered by your PR?

jesg:
>  is this issue covered by your PR?

yes.  ghostdriver will [set](https://github.com/detro/ghostdriver/blob/2.0.0/src/request_handlers/session_request_handler.js#L791) the domain if none is provided.

Vitallium:
Ok, closing this one then.

PR #14731, opened 2017-01-04
Update 2000-01-03-screen-capture.md

DanAtkinson:
Actually 49 lines now, but who's counting? :)


gillesdemey:
It would probably be better not to include the SLOCs in the README so we don't have to keep it up to date.

Vitallium:
I agree. Better to remove it.

DanAtkinson:
I've updated the pull to remove all SLOCs.

Vitallium:
Landed! Thank you!

PR #14824, opened 2017-02-16
Fix strict mode error

sj26:
When running this example I got the error message:

    ReferenceError: Strict mode forbids implicit creation of global property 'pageWidth'
   
      phantomjs://code/rasterize.js:29 in global code

This small change seems to fix it.


sj26:
Created an issue here:

14825

Vitallium:
Landed! Thank you!

PR #14838, opened 2017-02-19
remove phantom.args references

abrahamvarricatt:
phantom.args has been depreciated, system.args has been suggested as the alternative. 

This patch updates documentation to reflect the current situation.


ariya:
Thanks @abrahamvarricatt 👍

PR #14871, opened 2017-03-06
Fix for file uploads

raff:
Generate a native event (keyboard enter) instead of JS click on file selector button


ariya:
Thanks @raff! Do you have some tests for the code?

PR #14896, opened 2017-03-14
Add Homebrew instructions

ty-lg:


haroon-sheikh (approved):

ariya:
Unfortunately, that's not our policy. That page can only list installation methods that we officially support.
See #14896 for more details.

PR #14914, opened 2017-03-23
Update 2000-01-07-related-projects.md

merlinthemagic:
Added my project to the list of Related Projects. Used same language as PHP PhantomJS, since both projects act as a wrapper for PhantomJS.


pixiuPL:
duplication of scripts.

merlinthemagic:
I do not understand what you mean by "duplication of scripts.". Yes the text is the same, because the projects are very similar, both PHP and both provide a wrapper for PhantomJS. But the link text and link are different.

What are you looking for in order to merge?

Issue #14911, opened 2017-03-22
Bug with popup.settings.resourceTimeout and onResourceTimeout event


jefleponot:
1. Which version of PhantomJS are you using? Tip: run `phantomjs --version`.
1.9.7 - 2.1.1 - 2.5.0-dev
2. What steps will reproduce the problem?
  1.  open a page with a popup that calls a resource from a server that never reply any data

3. Which operating system are you using?
Linux : Centos / Ubuntu

4. Did you use binary PhantomJS or did you compile it from source?
Only Binary
5. Please provide any additional information below.

Works very well on the main page but never works on any popup page.

jefleponot:
@Vitallium @ariya 

Have you the same bug as me ? on phantomJS 2.1.1

```js
/*eslint strict:0*/
"use strict";
var server = require('webserver').create();
var service = server.listen(8092, function(request, response) {
    var path = request.url.split('/');
    response.statusCode = 200;
    switch (path[1]){
        case "index":
            response.setHeader('Content-type', 'text/html');
            response.write('<a href="/link">a link</a>');
            response.write('<form action="/form" method="POST"><input type="submit" /></form>');           
            response.write('<script src="/script"></script>');
            response.close(); 
            break;
        case "index2":
            response.setHeader('Content-type', 'text/html');
            response.write('<html><body><a href="/link">a link</a>');
            response.write('<form action="/form" method="POST"><input type="submit" /></form></body>');
            response.write('<script>window.open("http://localhost:8092/index","popup"); </script></html>');
            response.close(); 
            break;            
        case "form": case "link": case "":
            response.setHeader('Content-type', 'text/html');
            response.write('<a href="/link">a link</a>');
            response.write('<form action="/form" method="POST"><input type="submit" /></form>');   
            response.close();
            break;
        case "script": //never close connexion
            break;
        case "script2": //partial response never close connexion
            response.setHeader('Content-type', 'text/javascript');
            response.write('var a=2;');
            break;
        case "script3": //partial response never close connexion
            response.setHeader('Content-type', 'text/html');
            response.write('<html><script>window.open("http://localhost:8092/script") </script></html>');
            response.close();
            break;
        case "longScript": 
            response.setHeader('Content-type', 'text/html');
            response.write('<html><body><script>for(;;);</script></body></html>');
            response.close();
            break;
    }
});

var page = require('webpage').create();
var countResources = 0;
var idTest = 1;
var nbTest = 2;

[...]

PR #14924, opened 2017-03-28
add a resourceTimeout parameter to phantomjs from 2.1.1

jefleponot:
#14911

No event never trigger on popup opening. Because you can't precise a page.resourceTimeout before a "window.open event" that create the new popup.

And This timeout equals 0 by default. So It could be nice if we can fix it before running execution.
It will avoid crashs on no response services.

Regards


jefleponot:
@Vitallium / @ariya 
Could you please integrate this fix bug ?

jefleponot:
Is the 2.1.1 version is a Long Term Sable Executable ?

Vitallium:
@jefleponot Nop. We almost ready to release 2.5 version. With 2.5 release we will deprecate 2.1.

jefleponot:
@Vitallium 
It seams that the next release will have the same bug...

PR #14926, opened 2017-03-29
Make reading server.port return the correct value when port=0

### LINKED ISSUE #14925: Random port not reported correctly by webserver module
1. What version?
2.1.1

2. What steps will reproduce the problem?
    1. Get a copy of https://github.com/numtel/phantomjs-persistent-server/blob/master/src/phantom-server.js
    2. Change the last line to console.log(server.port, 'Ready.');
    3. Run it passing 0 (random port), and see that server.port returns 0 and not the port its bound to.

3. Which operating system are you using?
Linux

4. Did you use binary PhantomJS or did you compile it from source?
Binary

5. Please provide any additional information below.

### I wrote a patch I'm currently fashioning into a PR, which will reference this issue as requested in the CONTRIBUTING file.

Assigning a random port comes in handy if you're managing a bunch of phantom instances on one machine. Using 0 is the only race-condition-proof way of assigning a random port, as far as I know.

Here's a case where someone is working around the issue in a pretty heinous way that I didn't want to duplicate in my project:
* https://github.com/baudehlo/node-phantom-simple/blob/master/node-phantom-simple.js#L195-L229
* https://github.com/baudehlo/node-phantom-simple/issues/141

wcummings:
Tested w/ 2.1.1, building master now to test. webserver.cpp hasn't changed in a long time so it should work fine on master.

Fixes #14925


src/webserver.cpp
```diff
@@ -42,6 +42,7 @@
 #include <QUrl>
 #include <QVector>
 #include <QDebug>
+#include <sys/socket.h>
```
Vitallium: Windows does not have this header. You can remove it from here, `mongoose` already has it.

vitallium (dismissed):
In summary, I would like to not touch outdated mongoose source code, but move to a new version of it.

Vitallium:
I would like to not touch outdated mongoose source code, but move to a new version of it.

/cc @ariya @zackw @JamesMGreene

src/webserver.cpp
wcummings: :+1: good looks, this was included by accident

Vitallium: Yeah. I refreshed the page and this header has gone.

wcummings:
If it's just a matter of swapping out mongoose.{c,h} for new versions while preserving the WebServer interface, I can take a crack at it.

wcummings:
@Vitallium what do you think about using [node's http-parser](https://github.com/nodejs/http-parser) with Qt sockets?

PR #14929, opened 2017-03-31
add example

GQB1226:


_posts/api/webpage/method/2000-01-01-switch-to-frame.md
```diff
@@ -5,14 +5,22 @@ categories: api webpage webpage-method
 permalink: api/webpage/method/switch-to-frame.html
 ---
 
-`switchToFrame(frameName)` or `switchToFrame(framePosition)`
+if the page have othe frame, you can use `switchToFrame(frameName)` or `switchToFrame(framePosition)` to exchange it.Other similar methods are `switchToChildFrame()`,`switchToFocusedFrame`,`switchToMainFrame` and `switchToParentFrame`
```
haroon-sheikh: Looks like there is a typo with `othe` instead of `other`. Can you change it to following maybe to make it clearer:

```javascript
if the page has an other frame, you can use `switchToFrame(frameName)` or `switchToFrame(framePosition)` to switch to it.  Other options are `switchToChildFrame()`,`switchToFocusedFrame`,`switchToMainFrame` and `switchToParentFrame`
```

Issue #12192, opened 2014-05-02
Preserve fragment in URL on page redirect (HTTP status 302)


victor-homyakov:
All modern browsers preserve URL fragment on page redirect, but PhantomJS doesn't.

Test case: http://www.enhanceie.com/test/redir/fragment/redir-1.asp#Fragment

Tested browsers (Chrome 33+, IE10+, Firefox 28+) end with location.href == `http://www.enhanceie.com/test/redir/fragment/final.asp#Fragment`

But PhantomJS 1.9.7 ends with location.href == `http://www.enhanceie.com/test/redir/fragment/final.asp`

Therefore I cannot use PhantomJS to test my web application which uses redirection to the login page when user tries to open URL with fragment.

http://www.w3.org/TR/cuap#uri

> When a resource (URI1) has moved, an HTTP redirect can indicate its new location (URI2).
> 
> If URI1 has a fragment identifier #frag, then the new target that the user agent should be trying to reach would be URI2#frag. If URI2 already has a fragment identifier, then #frag must not be appended and the new target is URI2.
> 
> Wrong: Most current user agents do implement HTTP redirects but do not append the fragment identifier to the new URI, which generally confuses the user because they end up with the wrong resource.

http://blogs.msdn.com/b/ieinternals/archive/2011/05/16/url-fragments-and-redirects-anchor-hash-missing.aspx
http://stackoverflow.com/questions/5283395/url-hash-is-persisting-between-redirects

http://tools.ietf.org/html/draft-bos-http-redirect-00

> If the server returns a response code of 300 ("multiple choice"), 301 ("moved permanently"), 302 ("moved temporarily") or 303 ("see other"), and if the server also returns one or more URIs where the resource can be found, then the client SHOULD treat the new URIs as if the fragment identifier of the original URI was added at the end.

cschramm:
:+1:

SRGOM:
http://stackoverflow.com/questions/17982594/how-do-i-preserve-uri-fragment-in-safari-upon-redirect
http://stackoverflow.com/questions/2286402/url-fragment-and-302-redirects
https://bugs.webkit.org/show_bug.cgi?id=24175

QueShengyao:
The issue remains on 2.1.1 Win7 , phantomJS doesn't follow fragment  after redirection.(reproduced with @victor-homyakov 's example)

MahmoudIssaa:
Thanks for your Information!
[...]

PR #14941, opened 2017-04-10
attach fragment when redirect

QueShengyao:
attach fragment when redirect

if no fragment given in new url , attach the fragment of previous requsted url to new url

Issue: https://github.com/ariya/phantomjs/issues/12192


QueShengyao:
the changes are made based on 2.1 branch

pixiuPL:
@Vitallium what you think?

PR #15013, opened 2017-05-28
Get the QNetworkReply response body & Fix QWebSettings::WebSecurityEnabled bug

ibradypod:
Get the QNetworkReply response body at `onResourceRequested`.
Fix `WebSecurityEnabled` is not a member of `QWebSettings` bug.

### DIFF
--- a/src/networkaccessmanager.cpp
+++ b/src/networkaccessmanager.cpp
@@ -44,6 +44,10 @@
 #include "cookiejar.h"
 #include "networkaccessmanager.h"
 
+#include <private/qnetworkreplyhttpimpl_p.h>
+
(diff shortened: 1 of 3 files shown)


pixiuPL:
cc @Vitallium ??

Vitallium:
WebSecurity landed recently in QtWebkit, so this should be removed. And the code for getting body responses must be tested on AJAX queries. WebKit doesn't like when something intercepts them.

pixiuPL:
@Vitallium Close or merge?

Vitallium:
I think 'close'. Private methods are private for the reason. Anything can change inside of them. So, my point is to not use private methods.

PR #15025, opened 2017-06-09
Minor re-wording of doc and indentation fix

suganya-sk:
* Re-worded doc of `page.open` to reduce ambiguity
* Added new lines so the multiple possible invocations of `page.open` are shown in separate lines


Vitallium:
Landed! Thank you!

PR #15044, opened 2017-06-25
Stop using redirects on PhantomJS level

vitallium:
We should not use redirects on PhantomJS level because it breaks
internal network level inside WebKit.


tasktrello:
How can i handle redirect manually in case of JS redirect?

PR #15049, opened 2017-06-28
Add script to build for ppc64el

hamzy:
As documented here https://fedoraproject.org/wiki/User:Hamzy/phantomjs_ppc64le
This is a new script to build for the PowerPC architecture.  It could
still be merged with the x86_64 one, but the source code release version
numbers needs to be handled.


Vitallium:
I guess you want to merge it into 2.1 branch. Because this script is invalid for 2.5 version.

hamzy:
Good point... I'll close this one and submit another...

PR #15050, opened 2017-06-28
Add script to build for ppc64el

hamzy:
As documented here https://fedoraproject.org/wiki/User:Hamzy/phantomjs_ppc64le
This is a new script to build for the PowerPC architecture.  It could
still be merged with the x86_64 one, but the source code release version
numbers needs to be handled.


Vitallium:
Landed! Thank you!

Issue #15215, opened 2017-12-12
Error running examples/useragent.js


gverni:
1. Which version of PhantomJS are you using? Tip: run `phantomjs --version`.
2.1.1 
2. What steps will reproduce the problem?
  1.  Execute examples/useragent.js 
  2. The following error is shown 

`TypeError: null is not an object (evaluating 'document.getElementById('myagent').innerText')

  undefined:2
  :3
null`

3. Which operating system are you using?

Linux ubuntu 16.04 

4. Did you use binary PhantomJS or did you compile it from source?

Binary 

5. Please provide any additional information below.

The page at http://www.httpuseragent.org has probably changed and the "myagent" element does't exist anymore.

PR #15216, opened 2017-12-12
Fix for issue#15215

gverni:
In the page fetched from http://www.httpuseragent.org there is no DOM element with id "myagent". The userr agent is now in the DOM element with id qua. 

See issue https://github.com/ariya/phantomjs/issues/15215


pixiuPL:
can someone confirm this?

PR #15225, opened 2017-12-22
Strange chars removal

ghost:
Removed '=<branch>' chars from files.


pixiuPL:
@ariya @Vitallium what do you guys think?

PR #15308, opened 2018-02-06
Call correct method in example

sgrimm-sg:
The code example for makeTree is copied from the makeDirectory page and still calls makeDirectory instead of makeTree.


pixiuPL:
@sgrimm-sg good catch. Well done 👍

PR #15316, opened 2018-02-09
Updating GhostDriver

ghost:


jesg:
fwiw ghostdriver 2.1.0 only included changes to the java client.

pixiuPL:
@jesg but it was still worth updating....

PR #15323, opened 2018-02-14
Update 2000-01-01-open.md

alpar-t:
I couldn't really find more documentation on this, but the `utf8` in the example was not effective when dealing with a web page that has no character encoding specified. `ISO-8859-1` was used as a default and some characters corrupted in the process. `utf-8` solves this problem.


pixiuPL:
@atorok good catch. Thanks :)

Issue #13496, opened 2015-08-13
PhantomJS 2.0 and TLS 1.1 and 1.2


qaDream77:
So I've set my SSL protocol to be tlsv1.2 and yet when it tries to connect, the handshake fails.

2015-08-13T11:56:42 [DEBUG] 18 sslProtocol : "tlsv1.2"
2015-08-13T11:56:42 [DEBUG] 19 sslCiphers : "ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES128-SHA:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-RC4-SHA:ECDHE-RSA-RC4-SHA:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA:DHE-RSA-AES256-SHA:AES128-GCM-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:RC4-SHA:RC4-MD5"

Looking at the ciphers, there isn't a single TLS1.1 or TLS 1.2 cipher in there. Am i missing something or does it not support TLS (1.0 is disabled and isn't an option).

Any feedback would be greatly appreciated! Thank you.

Edit: when i do openssl ciphers in the command line, i have significantly more ciphers available to me to use. So how do i update the list of accessible ciphers for phantom?

Edit 2: Is phantom compiled with open ssl built in or does it used my systems current version of openssl?

zackw:
PhantomJS uses the system OpenSSL, and is supposed to support TLS1.1 and 1.2 if requested on the command line.  (Unfortunately, Qt currently doesn't give us a way to activate *all* of TLS 1.[012] simultaneously -- the default is [`QSsl::SecureProtocols`](http://doc.qt.io/qt-5.4/qssl.html#SslProtocol-enum) which is mapped to "SSL3.0 and TLS1.0" which is kinda wrong these days, but we don't have a better alternative short of patching qtbase ourselves.)

The default cipher list is
     // Default taken from Chromium 35.0.1916.153
    m_sslCiphers = ("ECDHE-ECDSA-AES128-GCM-SHA256"
                    ":ECDHE-RSA-AES128-GCM-SHA256"
                    ":DHE-RSA-AES128-GCM-SHA256"
[...]

PR #15332, opened 2018-02-23
Add support for specifying a minimum TLS version

robotfelix:
As well as supporting specific TLS versions (eg: v1.2), Qt supports specifying a minimum version (eg: v1.2 or later). This can be useful when the user wants to support multiple versions at once (for example, TLSv1.0 or later, but not SSLv3 or below), and support _**appears**_ to be trivial to add.

The possibility of using these additional Qt options was floated by @hughns issue #13496 in 2015 but did not receive any feedback.

As a note: I didn't touch https://github.com/ariya/phantomjs/blob/ddaaa09785d453e41554fd6133a2054c659a317f/src/config.cpp as it doesn't already mention support for TLSv1.1 and TLSv1.2. As I am new to this project, it would be good to know if the omission of these values was an oversight of a previous commit or is intentional?


pixiuPL:
@robotfelix ommision was not intentional. Good catch. Merging it.

PR #15350, opened 2018-03-07
An example config file for an angular application to configure the test report and mapping it to test reports(with an html file) to an html file.

sibStephen:
A reference file to configure the html reports, with following data.

1) Configuring the html reports with the html file using reporter object.
2) BeforeLaunch, onPrepare functions to perform an action at the initial stage before starting.


ariya:
Thanks @sibStephen for the example! But I don't think this script is intended to be executed with PhantomJS. Please note that PhantomJS != Node.js.

sibStephen:
Sure.

PR #15415, opened 2020-01-13
Replace NULL (or 0) with Q_NULLPTR

vitallium:
This:

* will make our code compiler-independent;
* will modernize our code with modern C++ standards.

This merge request also replaces several C style casts.

### DIFF
--- a/src/childprocess.cpp
+++ b/src/childprocess.cpp
@@ -92,11 +92,11 @@ qint64 ChildProcessContext::_write(const QString& chunk, const QString& encoding
     QTextCodec* codec = QTextCodec::codecForName(encoding.toLatin1());
 
     // If unavailable, attempt UTF-8 codec
-    if ((QTextCodec*)NULL == codec) {
+    if (!codec) {
         codec = QTextCodec::codecForName("UTF-8");
 
         // Don't even try to write if UTF-8 codec is unavailable
-        if ((QTextCodec*)NULL == codec) {
+        if (!codec) {
             return -1;
         }
     }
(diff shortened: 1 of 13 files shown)


ariya:
Looks @vitallium! Is that the CI error with coding style?

vitallium:
> Looks 👍 @vitallium! Is that the CI error with coding style?

Yeah, in `src/encoding.cpp`:
```diff
diff --git a/src/encoding.cpp b/src/encoding.cpp
index e0e07d331..5bd9e21b8 100644
--- a/src/encoding.cpp
+++ b/src/encoding.cpp
@@ -50,10 +50,10 @@ Encoding::Encoding(const QString& encoding)
 Encoding::~Encoding()
 {
     if (m_codec) {
-      // TODO: Encoding class does not inherit QObject, so
-      // we have to delete m_codec member manually;
-      m_codec = Q_NULLPTR;
-    }    
+        // TODO: Encoding class does not inherit QObject, so
+        // we have to delete m_codec member manually;
+        m_codec = Q_NULLPTR;
+    }
 }
 
 QString Encoding::decode(const QByteArray& bytes) const
```

ariya (approved):

ariya:
On a second thought, since likely we will only use C++11, isn't good to just stick with `nullptr`?

vitallium:
> On a second thought, since likely we will only use C++11, isn't good to just stick with `nullptr`?

I would say should we use more Qt specific things or write common C++ code without bringing Qt-related stuff, i.e. if we can write code without Qt then we just do it? If the answer is yes, then we should replace `Q_NULLPTR` with `nullptr`.

PR #15442, opened 2020-06-26
Support building for arm64 hosts

AGSaidi:
Since the wheezy repos aren't present anymore, also switch to using
the ubuntu:1604 docker image for compilation.

Compilation of an arm64 image works both on a native host or by docker using an arm64 base image and qemu

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -18,14 +18,26 @@ of the PhantomJS source repository:
 
 ```bash
  $ git clean -xfd .
- $ docker run -v $PWD:/src debian:wheezy /src/deploy/docker-build.sh
+ $ docker run -v $PWD:/src ubuntu:16.04 /src/deploy/docker-build.sh
 ```
 
 For the 32-bit version:
 
 ```bash
  $ git clean -xfd .
- $ docker run -v $PWD:/src tubia/debian:wheezy /src/deploy/docker-build.sh
+ $ docker run -v $PWD:/src i386/ubuntu:16.04 /src/deploy/docker-build.sh
+```
+
+For an arm64 version natively:
+```bash
+ $ git clean -xfd .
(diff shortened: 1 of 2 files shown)


ariya:
Thanks @AGSaidi for submitting this pull request!
Because there is no more maintenance being performed on the old 2.x versions, I'd rather leave the 2.xcode in its current state. I suggest creating a separate Git repository, e.g. `docker-build-phantomjs-arm`, that does the build using Docker for the said platforms.

AGSaidi:
I’m happy to take that approach @ariya but would you be willing to host the artifacts so anyone looking for an arm version would be able to easily find it? I’m also happy to help with getting future versions building.

ariya:
Hi @AGSaidi, if you make it so that the CI system (e.g. using GitHub Actions) runs the builds and saves the artifacts, then I will seriously consider pointing to the artifacts from the website. Thank you!

AGSaidi:
Thanks @ariya! I'll see what I can do with TravisCI since they support native hosted arm builds today

ariya:
Hi @AGSaidi, do you need that native hosted ARM since you can use Docker? BTW GitHub Actions is also a great alternative (than Travis CI) since it's well integrated with GitHub. See e.g. https://github.com/ariya/hello-c90/tree/master/.github/workflows.

AGSaidi:
@ariya, I don't think we want to try and cross compile everything and i'm worried about the build time for when running a arm64 docker container (under qemu) as i showed in the updated readme originally attached to this PR, but maybe it will build fast enough.

AGSaidi:
@ariya I have it building with travis-ci https://travis-ci.com/github/AGSaidi/phantomjs/builds/189875758

Apologies about the delay. I had to use a travis-ci VM to have enough disk space and that means using ubuntu:focal so the build is done in a container with xenial like my original PR. Would this be acceptable to you?

Issue #15414, opened 2020-01-13
Switch to Python 3


vitallium:
Given that Python 2 support [ended](https://pythonclock.org/) in 2019 and major Linux distributions will switch to Python 3 soon, for example, Python 2 and its related packages [will be removed](https://fedoraproject.org/wiki/Changes/RetirePython2) in Fedora 32, I suggest to upgrade our `./test/run-tests.py` to Python 3. WebKit is also [migrating](https://lists.webkit.org/pipermail/webkit-dev/2019-July/030724.html) to Python 3.

PR #15443, opened 2020-07-14
updated the run-test.py to work with python 3

anilreddy:
I have updated the run-test.py to work with python 3. Am able to run the scripts but are failing due to following error.
Can any one resolve this issue.

```python
ffFATAL: TypeError: cannot use a string pattern on a bytes-like object
## Traceback (most recent call last):
##   File "c:/Users/Anil Reddy/Downloads/phantomjs-master/test/run-tests.py", line 1057, in main    
##     sys.exit(runner.run_tests())
##   File "c:/Users/Anil Reddy/Downloads/phantomjs-master/test/run-tests.py", line 973, in run_tests
##     grp = self.run_test(test_script, tname)
##   File "c:/Users/Anil Reddy/Downloads/phantomjs-master/test/run-tests.py", line 950, in run_test 
##     grp.parse(rc, out, err)
##   File "c:/Users/Anil Reddy/Downloads/phantomjs-master/test/run-tests.py", line 643, in parse    
##     self.parse_tap(out, err)
##   File "c:/Users/Anil Reddy/Downloads/phantomjs-master/test/run-tests.py", line 655, in parse_tap
##     m = self.diag_r.match(line)
```

### DIFF
--- a/test/run-tests.py
+++ b/test/run-tests.py
@@ -4,18 +4,19 @@ import argparse
 import collections
 import errno
 import glob
-import imp
+import importlib
+import importlib.util
 import os
 import platform
 import posixpath
 import re
 import shlex
-import SimpleHTTPServer
+import http.server
 import socket
-import SocketServer
+import socketserver
 import ssl
 import string
-import cStringIO as StringIO
+from io import StringIO
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


anilreddy:
#15414

oddstr13:
That error is likely due to subprocess being opened in binary mode, and the output subsequently compared with a (unicode) string regex.

Fixing this could potentially be done by patting `encoding="utf-8"` to `subprocess.Popen`, but that will break backwards compatibility with `python < 3.6`.
https://github.com/anilreddy/phantomjs/blob/318de22aa67a9eab9320da3034e6154b8e4decac/test/run-tests.py#L189-L192

An alternative way would be to loop over the `stdout` and `stderr` result lists, and decode there (if needed).


I haven't dug into whether the output is expected to be binary data or textual data.

PR #15488, opened 2022-09-14
SRE-2900 Make binary available via GitHub

s01ipsist:
Commits the zipped binary so we can pull from GitHub rather than the "official" BitBucket store


s01ipsist:
Sorry, didn't mean to open upstream

